PR #364, opened 2016-02-01
Fixes layout problems. Updates footer to match Galter website and NW …

jstirnaman:
…branding.


phebal:
It looks good but doesn't seem like it fixed some of the problems on the home page.

The divider bar still does not extend all the way to the footer
middle_bar

This looks broken now:
feat-mobile-jason

Before it looked ok-is:
featured-stage

The two areas on the homepage are still not aligned:
not_alligned_now

Another example of thing not aligning:
jason-border-scroll
vs staging:
dash-ui-stage

Also the content border is now white and it should be dark gray like before.
Finally the mobile view scrolls horizontally

Note that I tested with my browser so it might look perfect on the phone. Let me know if that's so.

phebal:
The footer looks great btw

jstirnaman:
OK. I'll fix it. I prefer the white (or lighter gray?) border over dark gray, but your call.

phebal:
It was Violeta's preference to go with the dark-gray border. I personally don't care, feel free to talk to her about it. Thanks man.

jstirnaman:
I think I misunderstood which border. I see it now

jstirnaman:
@phebal Ready for review. Sorry for overlooking stuff last time. I just rebased it with your latest commits to master.

phebal:
Looks really good to me :+1: :star2:

PR #556, opened 2018-01-23
Scripts to load Library Notes and Arey book

phebal:
As requested by @sharpattack, here are the batch scripts for loading things into DH. The one for Arey might need to be updated a bit still.


sharpattack:
thank you @phebal!

PR #566, opened 2018-02-09
561 565 2017 sicw update instl perm form

sharpattack:


app/views/collections/_permission_form.html.erb
```diff
@@ -0,0 +1,121 @@
...
+  </div>
+
+  <!-- Share With -->
+  <div class="row">
+    <h3 class="col-sm-12">Share With <small>(optional)</small>
+      <span id="share_with_tooltip" class="h5"><%= share_with_help %></span>
+    </h3>
+  </div>
+  
+  <% unless f.object.institutional_collection %>
```
sharpattack: focus here @phebal when reviewing this file. this is mostly a copy of views/generic_files/_permission_form

spec/views/collections/_permissions.html.erb_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    }
+    
+    before do
+      instl_col.convert_to_institutional('institutional-shawshank')
+      assign(:collection, instl_col)
+      sign_in admin
+      render
+    end
+    
+    it 'hides user permissions field' do
```
phebal: I agree it should hide the permissions.

PR #577, opened 2018-02-19
allow twitterbot to crawl dh. fixes #573

### LINKED ISSUE #573: Change photo in Twitter when tweeting about an item in DigitalHub
The image displayed on twitter when tweeting about an item in DigitalHub is not very exciting. If possible change the photo to something more exciting or colorful? No ideas about what should take the current image's place...
digitalhub twitter

sharpattack:
Using twitter card validator, https://cards-dev.twitter.com/validator, our robots.txt disallow crawling to the path of the image.

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,3 +1,12 @@
+User-agent: Twitterbot
+Allow: /downloads
+Allow: /catalog
+Allow: /catalog?page*
+Disallow: /catalog?*
+Disallow: /catalog/*
+Disallow: /contact
+Disallow: /users
+
 User-agent: *
 Allow: /catalog
 Allow: /catalog?page*


public/robots.txt
```diff
@@ -1,3 +1,6 @@
+User-agent: Twitterbot
+Disallow:
```
phebal: Woudn't this let the bot crawl our catalog? This is not a good idea if so.

sharpattack: it should allow the bot to crawl everything, so yes the catalog too. will change it allow `/downloads/` which is where the image is, and follow the rest of the rules. will that be enough?

sharpattack:
@phebal you seemed uncertain, so I decided to double check and follow according to https://moz.com/learn/seo/robotstxt
please re-review

PR #579, opened 2018-02-20
Silence cron and switch email forwording to galter-is

phebal:
Tested on staging, works ok. Also tested the stats gathering task and it works fine.


config/schedule.rb
```diff
@@ -1,8 +1,8 @@
+set :output, '/home/deploy/cron.log'
```
sharpattack: good idea.

PR #583, opened 2018-02-22
Link DOIs to preferred resolver (fixes #582)

### LINKED ISSUE #582: Hyperlink DOIs to preferred resolver
Hello!

The DOI foundation recommends [this new, secure resolver](https://www.doi.org/doi_handbook/3_Resolution.html#3.8). There are several `.erb` files that link DOIs against the old one. If these can be savely changed, I'd be happy to send a PR.

Cheers :-)

katrinleinweber:
1st try to solver #582.

### DIFF
--- a/app/models/generic_file.rb
+++ b/app/models/generic_file.rb
@@ -122,7 +122,7 @@ class GenericFile < ActiveFedora::Base
     if doi.first.present?
       the_doi = doi.first.to_s.gsub('doi:', '')
       citation.gsub!(/ +$/, '')
-      citation << " <a href='http://dx.doi.org/#{the_doi}'>doi:#{the_doi}</a>"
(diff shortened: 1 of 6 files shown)


katrinleinweber:
I have a hunch that `gsub('doi:', '')` should be changed as well, or are you certain, that only that variant is possible as input? I've seen `DOI: ` and several different combinations of `https`, `dx.` & `doi.org` in the wild as well.

phebal:
Thanks Katrin. Tests pass for me and the new resolver works as expected from my tests. I'll go ahead and merge.

You're right about that `gsub`. It works OK for us because we generate DOI ourselves and moderate the user-entered metadata. I totally agree that this should be more robust and will open a ticket to address when we have time.

PR #586, opened 2018-03-02
Allow users to follow collections

phebal:
Also:
* Notifications about collection-type objects for the collections
* Notifications are authorized properly


config/routes.rb
```diff
@@ -39,10 +39,18 @@
   get '/iiif-api/generic_file/:id/annotation/:name', to: 'iiif_apis#annotation', as: 'iiif_apis_annotation'
   get '/iiif-api/generic_file/:id/list/:name', to: 'iiif_apis#list', as: 'iiif_apis_list'
 
+  #post '/collections/:id/follow', to: 'collections#follow', as: 'follow_collection'
+  #delete '/collections/:id/unfollow', to: 'collections#unfollow', as: 'unfollow_collection'
+
```
sharpattack: can i remove these @phebal ?

phebal: Please do.

app/views/users/_followed_collections_modal.html.erb
```diff
@@ -14,7 +14,11 @@
       <div class="modal-content">
         <div class="modal-body">
           <% if @user.all_followed_collections.count > 0 %>
-            <h3>Collections you follow</h3>
+            <% if current_user == @user %>
+              <h3>Collections you follow</h3>
+            <% else %>
+              <h3>Collections <%= @user.display_name %> follows</h3>
```
phebal: Thanks

config/initializers/event_job_initializer.rb
```diff
@@ -22,14 +22,28 @@ def collection_event(col)
...
         end
+        
         next unless user.can?(:read, col)
-        user.log_event(collection_event(col))
+
+        if parents.nil?
+          user.log_event(collection_event(col))
+        else 
+          parents.each do |parent|
+            if !parent.followers.include?(user)
```
phebal: This has a potential to spam. What if there were 5 parents? Would it log the event for `user` 5 times? I think to do what you're looking for, you need to get followers of all parent collections at once. Something like this would be enough to replace all the code you've added to this method and fix the problem:
```ruby
user.log_event(collection_event(col)) if !col.collections.map {|o| o.followers }.flatten.include?(user)
```

config/routes.rb
```diff
@@ -39,9 +39,6 @@
   get '/iiif-api/generic_file/:id/annotation/:name', to: 'iiif_apis#annotation', as: 'iiif_apis_annotation'
   get '/iiif-api/generic_file/:id/list/:name', to: 'iiif_apis#list', as: 'iiif_apis_list'
 
-  #post '/collections/:id/follow', to: 'collections#follow', as: 'follow_collection'
```
phebal: Thanks

spec/jobs/collection_upload_event_jobs_spec.rb
```diff
@@ -83,6 +84,35 @@
...
+
+      expect(job).to receive(:log_for_collection_follower).with(
+        job.collection
+      ).and_call_original
+      expect(job).to receive(:log_to_followers).and_call_original
+      job.run
+      
+      expect(follower4.events.length).to eq(1)
+    end
+  end
```
phebal: Needs multi-parent tests with scenarios where child's parents are not being followed by current user.

phebal:
@sharpattack did you fix that flaky test?

config/initializers/event_job_initializer.rb
sharpattack: i did not account for that. forgot that collections and generic files can belong to more than one collections. will update this. thanks for the code.

PR #592, opened 2018-03-23
Add altmetric donut

### LINKED ISSUE #574: Add Altmetric donut to DigitalHub item pages
Patty had a great idea (that we may have looked into in the past, but I can't find a ticket for it) to look into adding the Altmetric donut to our DigitalHub pages. 

Here's the information: 

https://www.altmetric.com/products/free-tools/institutional-repository-badges/

sharpattack:
Pending altmetric approval for our production and staging sites.
closes #574.

### DIFF
--- a/spec/views/generic_files/_show_actions.html.erb_spec.rb
+++ b/spec/views/generic_files/_show_actions.html.erb_spec.rb
@@ -1,0 +1,31 @@
+require 'rails_helper'
+
+describe 'generic_files/_show_actions.html.erb' do
+  let(:user) { create(:user) }
+  let(:gf) { make_generic_file(user, doi: ['test']) }
+  
+  before do
+    RSpec.configure do |config|
+      config.include Devise::TestHelpers, :type => :view
+    end
+    assign(:generic_file, gf)
+    sign_in(user)
+    expect(view).to receive(:can?).and_return(true).twice
+    render
+  end
+  
+  context 'altmetric' do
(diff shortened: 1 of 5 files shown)


spec/views/generic_files/_show_actions.html.erb_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    context 'with score of 0' do
+      it 'does not have link' do
+        expect(rendered).to_not have_selector('span.altmetric-embed > a')
+      end
+    end
+    
+    context 'with score > 0' do
+      it 'has a link to altmetric' do
+        skip "Check back after altmetric starts tracking and try with a gf on staging"
+        expect(rendered).to have_selector('span.altmetric-embed > a')
```
phebal: So should we wait for them to start tracking or is this safe to merge as is? I is deployed to staging?

sharpattack: hmm, let's wait until they start tracking the staging site.

sharpattack:
The folks at altmetric have just put tracking for our sites in place. This is deployed to staging,  but I don't think this is going to work there because the badge/donut requires a DOI, and the ones on staging are fake. Might have to put this in production and use real information to see it's use. 

What do you think @phebal ? It seems safe to me to put in production.

phebal:
Please do.


Thanks,

Piotr Hebal
________________________________

PR #596, opened 2018-03-27
Add bot-thwarting measures to contact form

### LINKED ISSUE #593: Try to prevent spam in the contact form
@jpastva  writes:

> I’m not sure if we have any trouble with our other web contact forms, but in the past month or so it seems like the spammers have ramped up their activity on the DigitalHub contact form, so I’m seeing 1-2 emails a day with junk, mostly linking to .ru email addresses. Is there anything we can do to minimize these? Wishful thinking?

phebal:
Fixes #593

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,5 +1,6 @@
 class ApplicationController < ActionController::Base
   before_action :configure_permitted_parameters, if: :devise_controller?
+  before_action :antispam_timestamp
   after_action :no_bot_crawl_on_staging
   rescue_from DeviseLdapAuthenticatable::LdapException do |exception|
     render :text => exception, :status => 500
@@ -18,6 +19,11 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
 
   private
+
+  config.antispam_threshold = 5
+  def antispam_timestamp
+    session['antispam_timestamp'] ||= Time.now
+  end
 
   def no_bot_crawl_on_staging
     if Rails.env.staging?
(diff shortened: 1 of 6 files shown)


app/controllers/application_controller.rb
```diff
@@ -19,6 +20,11 @@ class ApplicationController < ActionController::Base
 
   private
 
+  config.antispam_threshold = 5
```
sharpattack: this is set to 5 seconds? hopefully it's short/long enough. @jpastva let us know if you still get spam emails, as we could maybe bump this up. I can't see it taking anyone going straight /contact and speed typing a message in under 30 seconds, but maybe?

app/controllers/concerns/galtersufia/contact_form_controller_behavior.rb
```diff
@@ -7,6 +7,25 @@ module ContactFormControllerBehavior
...
     include Sufia::ContactFormControllerBehavior
 
+    included do
+      before_action :spam?, only: [:create]
+    end
+
+    def spam?
+      spam_msg = "potential spam detected for IP #{request.env['REMOTE_ADDR']}."
+      time_to_comment = Time.now - Time.parse(session['antispam_timestamp'])
+      if params[:contact_form][:customerDetail].present?
```
sharpattack: so a script/bot is supposed to fill this section of the form out? interesting solution.

phebal: All this is copied from what works for us on the website. Yes, the field is invisible to the user but visible to the bot that doesn't care about CSS-based properties.

app/controllers/application_controller.rb
jpastva: I'll let you know if they keep coming, thanks!

PR #626, opened 2018-09-05
Switch EZID to Datacite for DOI processing

phebal:
This code is already in production :(
I had to wait for them to move the prefix to finalize my tests.


kaars:
Thanks guys!

PR #647, opened 2019-02-20
original_publisher, publisher, and based_near Batch updates

### LINKED ISSUE #634: Batch metadata field updates
List of metadata fields to be changed in DigitalHub in batch if possible:

-All items with "Pediatric Neurology Briefs Publishers" in the Original Publisher field --> Add additional field to Digital Publisher with "Pediatric Neurology Briefs Publishers" and delete from Original Publisher
-All items in Science in Society Collection - delete "Science in Society" from Original Publisher
-All DigitalHub items - delete "Chicago, Illinois, United States" and delete "Chicago, IL, USA" from Location

It would be great if this could be done in next 2 months. Thanks!

sharpattack:
- closes #634

### DIFF
--- a/lib/batch/batch_metadata_updates_634.rb
+++ b/lib/batch/batch_metadata_updates_634.rb
@@ -1,0 +1,71 @@
+# batch updates for https://github.com/galterlibrary/digital-repository/issues/634
+
+# Updates number 1:
+# - All items with "Pediatric Neurology Briefs Publishers" in the Original 
+#   Publisher field --> Add additional field to Digital Publisher with "Pediatric 
+#   Neurology Briefs Publishers" and delete from Original Publisher
+
+pnb_publishers = "Pediatric Neurology Briefs Publishers"
+items_needing_original_publisher_updates = GenericFile.where("original_publisher" => pnb_publishers) + 
+                                           Collection.where("original_publisher" => pnb_publishers)
+
+items_needing_original_publisher_updates.each do |item|
+  item.publisher += [pnb_publishers]
+  item.original_publisher -= [pnb_publishers]
+  item.save!
+end
+
+
+# Updates number 2:
+# - All items in Science in Society Collection - delete "Science in Society" 
+#   from Original Publisher 
+
+sis_publisher = "Science in Society"
+
+def update_sis_original_publisher(id)
+  collection = Collection.find(id)
+
+  collection.members.each do |member|
+    if member.class == Collection 
+      update_sis_original_publisher(member.id)
+    else
+      member.original_publisher -= [sis_publisher]
+      member.save!
+    end
+  end
+
+  collection.original_publisher -= [sis_publisher]
+  collection.save!
+end
+
+sis_collection = "93c81706-47f8-49a1-88a2-3acbc971f4ed"
+update_sis_original_publisher(sis_collection)
+
+
+# Updates number 3:
+# - All DigitalHub items - delete "Chicago, Illinois, United States" 
+#   and delete "Chicago, IL, USA" from Location
+
+chicago_locations = ["Chicago, Illinois, United States", "Chicago, IL, USA"]
+items_needing_based_near_updates = GenericFile.where("based_near" => chicago_locations) +
+                                   Collection.where("based_near" => chicago_locations)
+
+items_needing_based_near_updates.each do |item| 
+  item.based_near -= chicago_locations
+  item.save!
+end
+
+
+# Updates number 4:
+# - All DigitalHub items with Digital Publisher value 
+#   "Galter Health Sciences Library, Feinberg School of Medicine, 
+#   Northwestern University" --> change to "Galter Health Sciences Library"
+
+ghsl_fsm_nu_publisher = ["Galter Health Sciences Library, Feinberg School of Medicine, Northwestern University"]
+items_needing_publisher_updates = GenericFile.where("publisher" => ghsl_fsm_nu_publisher) + 
+                                  Collection.where("publisher" => ghsl_fsm_nu_publisher)
+
+items_needing_publisher_updates.each do |item|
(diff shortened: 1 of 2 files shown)


lib/batch/batch_metadata_updates_634.rb
```diff
@@ -0,0 +1,73 @@
...
+#   - Northwestern University" --> change to "Galter Health Sciences Library"
+
+ghsl_fsm_nu_publisher = ["Galter Health Sciences Library, Feinberg School of Medicine, Northwestern University"]
+items_needing_publisher_updates = GenericFile.where("publisher" => ghsl_fsm_nu_publisher) + 
+                                  Collection.where("publisher" => ghsl_fsm_nu_publisher)
+
+items_needing_publisher_updates.each do |item|
+  item.publisher -= ghsl_fsm_nu_publisher
+  item.save
+end
```
fenekku: You were saying you had issues with the script not updating records. How did you test that this works?

lib/batch/batch_metadata_updates_634.rb
```diff
@@ -0,0 +1,73 @@
...
+#   - All items with "Pediatric Neurology Briefs Publishers" in the Original 
+#   - Publisher field --> Add additional field to Digital Publisher with "Pediatric 
+#   - Neurology Briefs Publishers" and delete from Original Publisher
+
+pnb_publishers = "Pediatric Neurology Briefs Publishers"
+items_needing_original_publisher_updates = GenericFile.where("original_publisher" => pnb_publishers) + 
+                                           Collection.where("original_publisher" => pnb_publishers)
+
+items_needing_original_publisher_updates.each do |item|
+  item.publisher += [pnb_publishers]
```
fenekku: Should we check that publisher contains `png_publishers` already? Is that a possibility?

lib/batch/batch_metadata_updates_634.rb
```diff
@@ -0,0 +1,73 @@
...
+# Updates number 2:
+#   - All items in Science in Society Collection - delete "Science in Society" 
+#   - from Original Publisher 
+
+sis_publisher = "Science in Society"
+
+def update_sis_original_publisher(id)
+  collection = Collection.find(id)
+
+  if !collection.members.empty?
```
fenekku: do you need this if you do an `each` in Ruby?

lib/batch/batch_metadata_updates_634.rb
```diff
@@ -0,0 +1,73 @@
...
+sis_collection = "93c81706-47f8-49a1-88a2-3acbc971f4ed"
+update_sis_original_publisher(sis_collection)
+
+
+# Updates number 3:
+#   - All DigitalHub items - delete "Chicago, Illinois, United States" 
+#   - and delete "Chicago, IL, USA" from Location
+
+chicago_locations = ["Chicago, Illinois, United States", "Chicago, IL, USA"]
+items_needing_based_near_updates = GenericFile.where("based_near" => chicago_locations) +
```
fenekku: What does `=>` in `"based_near" => chicago_locations` mean ? Does it mean contains "one or the other" or "both"? I am guessing it means "one or the other".

lib/batch/batch_metadata_updates_634.rb
sharpattack: So, `item.publisher` is an array of values. For some reason, using the shovel operator, `<<` was invalid. Using the `-=` or `+=` operators returns a new instance of the array, and was valid in saving.

lib/batch/batch_metadata_updates_634.rb
sharpattack: I checked and it won't save duplicate, so I think we're safe checking if it exists already.

lib/batch/batch_metadata_updates_634.rb
sharpattack: Maybe not. My first thought was to check if it had anything in the array.

lib/batch/batch_metadata_updates_634.rb
sharpattack: I did check this, and yes it returns "one or the other".

lib/batch/batch_metadata_updates_634.rb
sharpattack: But I should refactor this to `save!` to rescue instead of just returning false.

fenekku: I was also asking if you had forgotten to commit the tests you made for this? Or maybe you ran this on your local db?

lib/batch/batch_metadata_updates_634.rb
fenekku: Interesting, `item.publisher` is a `Set` or is filtered out of redundancies on save? How did you check?

lib/batch/batch_metadata_updates_634.rb
fenekku: Let's remove that check. If `collection.members` is an empty array the `each` won't be executed.

lib/batch/batch_metadata_updates_634.rb
sharpattack: I checked this by manually testing it. I will look to see if there is a test that checks this scenario, and add one if not.

lib/batch/batch_metadata_updates_634.rb
sharpattack: okay, sounds fine to me.

lib/batch/batch_metadata_updates_634.rb
sharpattack: There already is a test that checks updating these values. It took me time to realize that none of these tests were using the `<<` operator, and that it appeared to be in a specific format that was valid for updating. That was when I tried the specific way of updating, which is using `-=` , `+=`, or just `=`.

Issue #635, opened 2018-11-29
Generate monthly exports of DigitalHub metadata


jpastva:
Run a report at the beginning of each month that includes all descriptive metadata for the prior month's new deposits (formatted according to previous export requests). For example, on 12/1/2018, export metadata for all items deposited from 11/1/2018-11/30/2018.

File of exports can be emailed, or loaded into Box (https://northwestern.box.com/s/ezk13jcoad7oqr9fue4nkhsimzk8r2bf)

PR #649, opened 2019-02-26
closes #644 - add metadata extraction script

### LINKED ISSUE #644: Move metadata extraction script to this project
Currently, the script is just living on the server in a folder. It should be a part of this project's repository, with instructions to run it. 

Related to #536

sharpattack:
- Script was just living on the server, and should be a part of
  the code base so we do not lose it. We will need to come up
  with a different solution for these requests though, see #635

### DIFF
--- a/lib/batch/subject.csv.rb
+++ b/lib/batch/subject.csv.rb
@@ -1,0 +1,24 @@
+csv = CSV.open('/home/deploy/all_gf_subject.csv', 'wb')
+terms =  GalterGenericFilePresenter.terms + [:depositor, :date_uploaded, :date_modified, :read_access_group]
+
+csv << terms.map(&:to_s).map { |term| 
+  I18n.translate(:simple_form)[:labels][:generic_file][term] || term.to_s.titleize 
+}
+
+ActiveFedora::SolrService.query('has_model_ssim:GenericFile', { rows: 99999 }).each.with_index do |gf, idx|
+  gfmod = gf.each.map {|k, v| [k.gsub(/_[a-z]+\z/, ''), v] }.to_h
+  csv << terms.map {|term|
+    t = term.to_s
+    (gfmod[t].is_a?(Array) ? gfmod[t].reject(&:blank?).compact.join(' ; ') : gfmod[t]).to_s
+  }
+  puts "Index: #{idx}" if idx % 100 == 0
+end
+ActiveFedora::SolrService.query('has_model_ssim:Page', { rows: 99999 }).each.with_index do |gf, idx|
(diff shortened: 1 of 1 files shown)


lib/batch/subject.csv.rb
```diff
@@ -0,0 +1,22 @@
+csv = CSV.open('/home/deploy/all_gf_subject.csv', 'wb')
+terms =  GalterGenericFilePresenter.terms + [:depositor, :date_uploaded, :date_modified, :read_access_group]
+
+csv << terms.map(&:to_s).map {|t| I18n.t(:simple_form)[:labels][:generic_file][t] || t.to_s.titleize }
```
fenekku: Let's change `|t|` to `|term|` or `|e|`? `t` is used for `I18n.translate` and that is unnecessarily confusing.

lib/batch/subject.csv.rb
```diff
@@ -0,0 +1,22 @@
...
+
+ActiveFedora::SolrService.query('has_model_ssim:GenericFile', { rows: 99999 }).each.with_index do |gf, idx|
+  gfmod = gf.each.map {|k, v| [k.gsub(/_[a-z]+\z/, ''), v] }.to_h
+  csv << terms.map {|term|
+    t = term.to_s
+    (gfmod[t].is_a?(Array) ? gfmod[t].reject(&:blank?).compact.join(' ; ') : gfmod[t]).to_s
+  }
+  puts "Index: #{idx}" if idx % 100 == 0
+end
+ActiveFedora::SolrService.query('has_model_ssim:Page', { rows: 99999 }).each.with_index do |gf, idx|
```
fenekku: Does this limit the total rows or paginates them? If the former, we know for sure that we have less than 100,000 records?

lib/batch/subject.csv.rb
```diff
@@ -0,0 +1,22 @@
+csv = CSV.open('/home/deploy/all_gf_subject.csv', 'wb')
```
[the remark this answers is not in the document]
sharpattack: I think so. We can change the file name to anything we want though. Just got to make sure we make the same changes anywhere else that needs it.

lib/batch/subject.csv.rb
sharpattack: Ha, it is. I will change to `term` and `I18n.translate`.

lib/batch/subject.csv.rb
sharpattack: It limits the rows. Without checking somehow before every time we run this script, we will not know for sure if we have less than this parameter. There is a plan to come up with a different solution to handle this, so we can address this then.

fenekku:
Great thanks for the answers!

PR #657, opened 2019-05-01
Disable Original Publisher field

### LINKED ISSUE #655: Disable Original Publisher field
Disable Original Publisher field, and transfer values to Digital Publisher field. As of the creation of this ticket, there were no values in the Original Publisher field. Ideally this merge will happen ASAP to prevent any further use of the Original Publisher field.

sharpattack:
- transfer the values to (Digital) Publisher field. closes #655

### DIFF
--- a/lib/batch/batch_metadata_updates_655.rb
+++ b/lib/batch/batch_metadata_updates_655.rb
@@ -1,6 +1,6 @@
 # Transfer metadata values from Original Publisher to Digital Publisher field
 
-# below is too slow
+# below is too slow, overloads the memory
 #items_needing_transfer = GenericFile.all.reject { |gf| 
 #                           gf.original_publisher == []
 #                         }
@@ -22,6 +22,6 @@ solr_items_w_original_publisher_value.each do |solr_item|
     end
   end
 
-  item.original_publisher -= item.original_publisher
+  item.original_publisher = []
   item.save!
 end
--- a/lib/batch/batch_metadata_updates_655.rb
+++ b/lib/batch/batch_metadata_updates_655.rb
@@ -1,0 +1,27 @@
+# Transfer metadata values from Original Publisher to Digital Publisher field
+
+# below is too slow
+#items_needing_transfer = GenericFile.all.reject { |gf| 
+#                           gf.original_publisher == []
+#                         }
+
+solr_items_w_original_publisher_value = ActiveFedora::SolrService.query(
+                                     'original_publisher_tesim:*', 
+                                     { rows: 99999 }
+                                   )
+
+solr_items_w_original_publisher_value.each do |solr_item|
+  # could be either Collection or GenericFile, so we convert the string to
+  # class to get the object by id
+  constant_item = solr_item['has_model_ssim'].first.constantize
+  item = constant_item.find(solr_item['id'])
+
+  item.original_publisher.each do |op|
+    if !item.publisher.include?(op)
+      item.publisher += [op]
+    end
+  end
+
+  item.original_publisher -= item.original_publisher
+  item.save!
+end
--- a/app/presenters/galter_collection_presenter.rb
+++ b/app/presenters/galter_collection_presenter.rb
@@ -17,7 +17,6 @@ class GalterCollectionPresenter < Sufia::CollectionPresenter
       :bibliographic_citation,
       :related_url,
       :publisher,
-      :original_publisher,
       :date_created,
       :identifier,
       :language,
--- a/app/presenters/galter_generic_file_presenter.rb
+++ b/app/presenters/galter_generic_file_presenter.rb
@@ -11,7 +11,6 @@ class GalterGenericFilePresenter < Sufia::GenericFilePresenter
     :bibliographic_citation,
     :related_url,
     :publisher,
-    :original_publisher,
     :date_created,
     :identifier,
     :language,


lib/batch/batch_metadata_updates_655.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  item.original_publisher.each do |op|
+    if !item.publisher.include?(op)
+      item.publisher += [op]
+    end
+  end
+
+  item.original_publisher -= item.original_publisher
+  item.save!
+end
```
fenekku: One-time scripts don't *necessarily* warrant tests, but this kind might. It would help:
- to make sure we are not altering something we don't want to
- to get faster feedback on what works and what doesn't
- to convince me this works (don't know the structure of the index, so don't know if this works well)

I would emphasize the 2nd point: testing this script on a complete index is very time consuming. Not only for the script, but for you: you need to reload the full index if you made a mistake (you maybe are running this on a remote machine which adds time...) Using a smaller one might help you iterate faster.

lib/batch/batch_metadata_updates_655.rb
```diff
@@ -0,0 +1,27 @@
...
+  constant_item = solr_item['has_model_ssim'].first.constantize
+  item = constant_item.find(solr_item['id'])
+
+  item.original_publisher.each do |op|
+    if !item.publisher.include?(op)
+      item.publisher += [op]
+    end
+  end
+
+  item.original_publisher -= item.original_publisher
```
fenekku: We can't just assign `[]` here?

lib/batch/batch_metadata_updates_655.rb
```diff
@@ -0,0 +1,27 @@
...
+solr_items_w_original_publisher_value = ActiveFedora::SolrService.query(
+                                     'original_publisher_tesim:*', 
+                                     { rows: 99999 }
+                                   )
+
+solr_items_w_original_publisher_value.each do |solr_item|
+  # could be either Collection or GenericFile, so we convert the string to
+  # class to get the object by id
+  constant_item = solr_item['has_model_ssim'].first.constantize
+  item = constant_item.find(solr_item['id'])
```
fenekku: Is this a DB call or a Solr call? How slow is this script? Have you looked into Solr/DB's bulk update?

lib/batch/batch_metadata_updates_655.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  item.original_publisher.each do |op|
+    if !item.publisher.include?(op)
+      item.publisher += [op]
+    end
+  end
+
+  item.original_publisher -= item.original_publisher
+  item.save!
```
fenekku: Does this `save!` reindex? If entries are in DB *and* index we want to make sure both are in sync.

lib/batch/batch_metadata_updates_655.rb
```diff
@@ -0,0 +1,27 @@
+# Transfer metadata values from Original Publisher to Digital Publisher field
+
+# below is too slow
+#items_needing_transfer = GenericFile.all.reject { |gf| 
+#                           gf.original_publisher == []
+#                         }
+
+solr_items_w_original_publisher_value = ActiveFedora::SolrService.query(
+                                     'original_publisher_tesim:*',
```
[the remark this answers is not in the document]
sharpattack: It's a solr dynamic field. 
From solr docs: https://lucene.apache.org/solr/guide/7_0/dynamic-fields.html
In our config: https://github.com/galterlibrary/digital-repository/blob/master/solr_conf/conf/schema.xml#L41

lib/batch/batch_metadata_updates_655.rb
sharpattack: We can, and it is probably a better idea. I had trouble in the past with other ways of saving changes to other fields, but this seems to work.

lib/batch/batch_metadata_updates_655.rb
sharpattack: Yes, `save!` reindexes, and keeps Fedora and index objects in sync.

lib/batch/batch_metadata_updates_655.rb
sharpattack: It's an ActiveFedora call, and does not have a bulk update. This isn't slow, but it's not super fast. Querying the index was instant, and doing a `.find()` for each GenericFile or Collection was faster than getting all the ActiveFedora objects, then rejecting ones with empty `original_publisher` fields. 

The slow comment I made was in reference to `GenericFile.all.reject`

lib/batch/batch_metadata_updates_655.rb
sharpattack: Sure, I agree we should test when needed. But maybe answering some concerns may prove it's not necessary for this one. 

- I know we won't be altering the wrong items, because Original Publisher was custom added to only the two; Collection and GenericFile
- It's not needed to know the structure of the index. We only query it because it's much faster than ActiveFedora(these are not DB objects, but Fedora objects) to get the objects we want to update. 
- I have been testing this manually on my local machine(for what it's worth), so testing against my development index works. 

If these are still not convincing enough, I'll look into how to test this.

lib/batch/batch_metadata_updates_655.rb
fenekku: > It's an ActiveFedora call

... which is a DB or a Solr call ;) ? I *think* it is a DB call. I *think* ActiveFedora provides an ActiveSupport like ORM over Fedora's entries. But then why would this approach be any faster? Unless `.all` loads *everything* in memory and this chokes your system.

lib/batch/batch_metadata_updates_655.rb
fenekku: In pragmatic terms, my concern is that using the full index/db as the basis to iterate development might be too slow.
Given that DigitalHub is in maintenance mode, it is not too bad if we don't have a test index. I just want to highlight the potential stakes:

Unless you always get it right on the first pass of your code, I only see how slow and painful the experience must be: write code, run on db/index, oups doesn't work, reload full db and index, write code, oups, full reload... 
This is also why I am wondering how slow/fast is the running code (order of magnitude), because that will also hamper this flow. Finally, this is at least the second time something like this is run on DigitalHub. Having a nice small setup for this would save time every one of those instances.

Just something to keep in mind in the future. You can merge this.

lib/batch/batch_metadata_updates_655.rb
sharpattack: Yes, you are right about all of this. In my mind I wasn't sure to call Fedora's entries a DB object, though ActiveFedora is based on ActiveSupport like ORM for Fedora. With that said, the `.all` does overload the memory, which is why this approach is better.

lib/batch/batch_metadata_updates_655.rb
sharpattack: Ahh, I understand now. Yes, I can see how this can be a headache without tests, and I see the value now. As you stated earlier, one time scripts don't always need it, and because of that it has never crossed my mind. But I will definitely try that route for the next one(probably as soon as the next PR). Thanks for raising these concerns.

PR #659, opened 2019-05-06
Release 1.15.0-disable_Original_Publisher-update_and_institutionalize_scripts

sharpattack:
Batch update for Publisher field script needs to be run. The rest were manually uploaded to the server and already run.

### DIFF
--- a/app/presenters/galter_collection_presenter.rb
+++ b/app/presenters/galter_collection_presenter.rb
@@ -17,7 +17,6 @@ class GalterCollectionPresenter < Sufia::CollectionPresenter
       :bibliographic_citation,
       :related_url,
       :publisher,
-      :original_publisher,
       :date_created,
       :identifier,
       :language,
--- a/app/presenters/galter_generic_file_presenter.rb
+++ b/app/presenters/galter_generic_file_presenter.rb
@@ -11,7 +11,6 @@ class GalterGenericFilePresenter < Sufia::GenericFilePresenter
     :bibliographic_citation,
     :related_url,
     :publisher,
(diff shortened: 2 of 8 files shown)


sharpattack:
This is on staging, and `batch_metadata_updates_655.rb` has been successfully run.

jpastva:
@fenekku _Galter Health Sciences Library_ and _Galter Health Sciences Library & Learning Center_ should be kept separate. The shorter name was our official name in use at the time of the launch of DigitalHub, and the longer one is the result of a name change and is currently in use. So both are correct.

sharpattack:
Thanks for confirming @jpastva 
@fenekku the change to "Publisher" will happen in an upcoming PR.

fenekku:
@jpastva thanks for the information. Why isn't the name changed retroactively? Both Galter Health Sciences Library and Galter Health Sciences Library & Learning Center refer to the same institution. Versioning could help keep history here. For DigitalHub it is fine. I am thinking mostly about the new system and how it is a minor annoyance + visually messy to filter by a publisher, just to realize there are variations on that publisher.

@sharpattack ok for the Publisher change.

jpastva:
@fenekku This is standard cataloging practice, to record the name of the publisher at the time a resource was published. I agree that this is somewhat annoying for filtering, but changing the practice would open a whole other can of worms (database maintenance for any other publisher changes, the potential for misleading information if a name change was more dramatic, etc.). Typically name authority records are useful in sorting out relationships when names change, though I'm not sure how they could be leveraged in the new system.

PR #661, opened 2019-05-06
Remove Original Publisher

### LINKED ISSUE #656: Remove Original Publisher
Undo #384 

Depends on #655 

And rename "Digital Publisher" to "Publisher"

sharpattack:
- and rename Digital Publisher to Publisher. closes #656


spec/controllers/collections_controller_spec.rb
```diff
@@ -137,7 +133,6 @@
       expect(assigns(:collection).subject_geographic).to eq(['geo'])
       expect(assigns(:collection).subject_name).to eq(['subjn'])
       expect(assigns(:collection).multi_page).to eq(true)
-      expect(assigns(:collection).original_publisher).to eq(['opub'])
       expect(assigns(:collection).private_note).to eq(['note'])
```
fenekku: `digital_publisher` was not tested against collection in the first place? or is it under the `digital_origin` label?

sharpattack: Doing some digging, I found there was no `digital_publisher` attribute, it was always `publisher`. The "Digital Publisher" was custom at the view level to display for the `publisher` attribute. This block tests "custom added" attributes, which`original_publisher` was. While the "Digital Publisher" was custom, this is at the view level, and the attribute was not added.

sharpattack: Actually, `publisher` will be tested in an upcoming PR with it prepopulated.

fenekku: My approval holds and I will look at the `publisher` stuff in the next PR. Thanks.

PR #667, opened 2019-05-31
Release 1.16.0-remove_Original_Publisher-prefill_Publisher

sharpattack:
- prefill Publisher field with "DigitalHub. Galter Health Sciences Library & Learning Center"
- update Publisher fields with "Galter Health Sciences Library" and "Galter Health Sciences Library & Learning Center" to be prepended with "DigitalHub. "
- add "DigitalHub. Galter Health Sciences Library & Learning Center" to non-private empty Publisher fields


fenekku:
@sharpattack Where is this already deployed? Should I expect the publisher field to be prefilled when I generate a citation?

sharpattack:
@fenekku it wasn't deployed, it is now. 

You should expect the Publisher field to be prefilled when you upload a new file or create a new collection. I have not checked how the generated citation looks, but if you see it's not, let me know.

fenekku:
Works for me :+1: 

I just realized that since the update script doesn't affect private records, when those are made public, the publisher won't be updated resulting in public entries with empty publishers. I wouldn't bother fixing that though given our context.

PR #673, opened 2019-06-27
[#672] Handle Abusive Requests

### LINKED ISSUE #672: Block and throttle abusive requests
Preventive measures from potential threats such as bad bots and (D)DoS events.

sharpattack:
Use rack-attack to throttle requests. Based on recent events, we
track subnets and set limits based on:
- if a user is signed in or not
- if the request path is to /catalog(?|/)

NU networks are whitelisted, and known abusive subnets are handled in
the apache configs. closes #672

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -1,0 +1,40 @@
+class Rack::Attack
+  # cache store
+  Rack::Attack.cache.store = ActiveSupport::Cache::MemoryStore.new
+
+  # whitelist NU network
+  safelist_ip("165.124.0.0/16")
+  safelist_ip("129.105.0.0/16")
+
+  # Always allow requests from localhost
+  # (blocklist & throttles are skipped)
+  Rack::Attack.safelist('allow from localhost') do |req|
+    '127.0.0.1' == req.ip || '::1' == req.ip
+  end
+
+  # These include some requests by the application, and not by the user
+  Rack::Attack.safelist('ignorable paths') do |req|
+    req.fullpath.start_with?('/assets',
+                             '/users/notifications_number', 
+                             '/catalog/facet/tags_sim.json',
+                             '/downloads/')
+  end
+
+  catalog_limit_proc = proc { |req| req.env['warden'].user.blank? ? 10 : 500 }
+  # Throttle /catalog by subnet
+  # 500/6mins for authenticated, 10/6mins for unauthenticated
+  # Key: "rack::attack:#{Time.now.to_i/:period}:catalog/subnet:#{subnet of req.ip}"
+  throttle('catalog/subnet', limit: catalog_limit_proc, period: 6.minutes) do |req|
+    if req.fullpath.start_with?('/catalog?', '/catalog/') && !req.fullpath.start_with?('/catalog?page')
+      req.ip.slice(0..req.ip.rindex("."))
+    end
+  end
+
+  all_limit_proc = proc { |req| req.env['warden'].user.blank? ? 100 : 1000 }
+  # Throttle all requests by subnet
+  # 1000/6mins for for authenticated, 100/6mins for unauthenticated
+  # Key: "rack::attack:#{Time.now.to_i/:period}:req/subnet:#{subnet of req.ip}"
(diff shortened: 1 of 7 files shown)


config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,40 @@
...
+                             '/downloads/')
+  end
+
+  catalog_limit_proc = proc { |req| req.env['warden'].user.blank? ? 10 : 500 }
+  # Throttle /catalog by subnet
+  # 500/6mins for authenticated, 10/6mins for unauthenticated
+  # Key: "rack::attack:#{Time.now.to_i/:period}:catalog/subnet:#{subnet of req.ip}"
+  throttle('catalog/subnet', limit: catalog_limit_proc, period: 6.minutes) do |req|
+    if req.fullpath.start_with?('/catalog?', '/catalog/') && !req.fullpath.start_with?('/catalog?page')
+      req.ip.slice(0..req.ip.rindex("."))
```
fenekku: Let's document in comments that we are throttling every request based on their **/24** subnet. It will be interesting to see how users react to this in practice; throttling by subnet like this may result in situations where a user is poisoning the well for everyone else. (which is why this should not be done for an API)

spec/requests/rack_attack_spec.rb
```diff
@@ -0,0 +1,242 @@
...
+
+      it 'does not track the ip' do
+        expect(request.env['rack.attack.matched']).to eq('allow from localhost')
+        expect(request.env['rack.attack.match_type']).to eq(:safelist)
+        expect(request.env['rack.attack.match_data']).to_not be_present
+        expect(request.env['REMOTE_ADDR']).to eq('::1')
+      end
+    end
+
+    describe 'NU network 165.214.0.0/16' do
```
fenekku: typo: `165.214.0.0/16` -> `165.124.0.0/16`

spec/requests/rack_attack_spec.rb
```diff
@@ -0,0 +1,242 @@
...
+          data = { :count => 1, :limit => 100, :period => @period.to_i, :epoch_time => @epoch_time }
+          expect(request.env['rack.attack.throttle_data']['req/subnet']).to eq(data)
+          expect(request.env['REMOTE_ADDR']).to eq('1.2.3.4')
+        end
+
+        it 'suppresses the subnet' do
+          key = "rack::attack:#{(Time.now.to_i/@period).to_i}:req/subnet:1.2.3."
+          Rack::Attack.cache.store.increment(key, 99)
+          @epoch_time = Time.now.to_i
+          get '/about', {}, 'REMOTE_ADDR' => '1.2.3.5'
```
fenekku: minor: add new line above `get` and remove new line between `data` and `expect`

config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,40 @@
...
+                             '/users/notifications_number', 
+                             '/catalog/facet/tags_sim.json',
+                             '/downloads/')
+  end
+
+  catalog_limit_proc = proc { |req| req.env['warden'].user.blank? ? 10 : 500 }
+  # Throttle /catalog by subnet
+  # 500/6mins for authenticated, 10/6mins for unauthenticated
+  # Key: "rack::attack:#{Time.now.to_i/:period}:catalog/subnet:#{subnet of req.ip}"
+  throttle('catalog/subnet', limit: catalog_limit_proc, period: 6.minutes) do |req|
```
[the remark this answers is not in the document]
sharpattack: I will remove the "Key" comment. It was added following examples in rack-attack's page. 

Same about the 6 min period setting: I did not see any examples that had the time period as high as an hour. I can see users hitting the 10/6min limit, so bumping it to 1 hour may be more beneficial.

config/initializers/rack_attack.rb
sharpattack: `poisoning the well` - great analogy.

spec/requests/rack_attack_spec.rb
sharpattack: Nice catch.

spec/requests/rack_attack_spec.rb
sharpattack: Done.

Issue #696, opened 2019-10-29
Make throttle rules stricter


sharpattack:
The DDoS-like event is back. The throttle rules are not enough, and we will try to making the rules stricter, at least to the "/catalog" path. 

See #672 for more background.

sharpattack:
Going to go with exponential backoff https://github.com/kickstarter/rack-attack/wiki/Advanced-Configuration#exponential-backoff

And add custom throttle response to inform users to sign in for "Better user experience".
https://github.com/kickstarter/rack-attack/wiki/Example-Configuration

sharpattack:
This is done, forgot to close in commit 7ea3633

PR #697, opened 2019-10-30
[#696] Exponentially throttle requests to /catalog, and more

sharpattack:
- Exponentially throttle unauthenticated user requests to /catalog(?|/)
  path.
- Custom throttled response to inform users to log in.
- Do not throttle authenticated/logged in users.

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -20,25 +20,48 @@ class Rack::Attack
                              '/downloads/')
   end
 
+  # Allow authenticated user
+  Rack::Attack.safelist('authenticated user') do |req|
+    !req.env['warden'].user.blank?
+  end
+
   # Based on occurrences during the week of June 16, 2019, DigitalHub
   # suffered a DDoS-like event. Inspection of logs during the event showed
   # +10k requests per hour to digitalhub, from thousands of different ips.
   # Throttling by ip would prove ineffective in this case. Though in the
   # hundreds, the ips showed a subnet-like pattern. The following throttle
   # rules, tracking by subnet, are based off this event.
-  catalog_limit_proc = proc { |req| req.env['warden'].user.blank? ? 100 : 5000 }
+
   # Throttle /catalog by subnet
-  # 5000/hr for authenticated, 100/hr for unauthenticated
-  throttle('catalog/subnet', limit: catalog_limit_proc, period: 1.hour) do |req|
-    if req.fullpath.start_with?('/catalog?', '/catalog/') && !req.fullpath.start_with?('/catalog?page')
-      req.ip.slice(0..req.ip.rindex("."))
+  # Exponential Backoff for unauthenticated user
+  #   Level 6: 96 requests per 64 minutes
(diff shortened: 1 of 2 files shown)


spec/requests/rack_attack_spec.rb
```diff
@@ -98,59 +105,156 @@
...
+        it 'suppresses the subnet with custom response' do
           key = "rack::attack:#{(Time.now.to_i/@period).to_i}:req/subnet:1.2.3."
           Rack::Attack.cache.store.increment(key, 999)
           @epoch_time = Time.now.to_i
 
           get '/about', {}, 'REMOTE_ADDR' => '1.2.3.5'
           data = { :count => 1001, :limit => 1000,
                    :period => @period.to_i, :epoch_time => @epoch_time }
-          expect(response.status).to eq(429)
-          expect(response.headers['Retry-After']).to eq(@period.to_s)
```
fenekku: minor: we don't want to keep that header?

config/initializers/rack_attack.rb
```diff
@@ -20,25 +20,48 @@ class Rack::Attack
...
+  #   Level 6: 96 requests per 64 minutes
+  #   Level 7: 112 requests per 128 minutes
+  #   Level 8: 128 requests per 256 minutes
+  #   Level 9: 144 requests per 512 minutes
+  #   Level 10: 160 requests per 1024 minutes
+  (6..10).each do |level|
+    # We set a strict limit for the catalog path because facets/filters consume
+    # a lot of resources, hence slowing down digitalhub. Navigating around the
+    # '/catalog?page' path is fine.
+    throttle("catalog/subnet/#{level}", :limit => (16 * level), :period => (2 ** level).minutes) do |req|
```
fenekku: What is the meaning of `level` here?

The exponential part is done by us. Does `throttle` do the backoff part?

sharpattack: From levels 6-10, we set different throttle rules for each. For each level, we decrease the requests rate limit, and increase the backoff period. Does that help? I can explain in person if easier?

config/initializers/rack_attack.rb
```diff
@@ -20,25 +20,48 @@ class Rack::Attack
...
     req.ip.slice(0..req.ip.rindex("."))
   end
+
+  # Custom throttle response
+  self.throttled_response = lambda do |env|
+    [
+      503, # status
+      {}, # headers
+      ['Looks like you have reached your limit. ',
+       'Please log in to get the full experience.'] # body
```
sharpattack: @fenekku I was hoping for a comment on this message 😁, what do you think?

spec/requests/rack_attack_spec.rb
fenekku: Either way is fine. Just making sure it wasn't an oversight. Since we return 503 now, we can omit it.

config/initializers/rack_attack.rb
fenekku: That works :heavy_check_mark:

PR #708, opened 2019-12-11
703 center for data science and informatics, 707 template collection script

sharpattack:
The currently existing "2016" collections that need to be institutionalized, will be done in a separate task. 

The subcollections "Biomedical Data Science Day", "2020", and "Chicago Biomedical Informatics and Data Science Jam" will be created after using some of the collections created from the template collection script.


sharpattack:
After an approved review, I will deploy to staging, run each script, and report any errors.

lib/scripts/create_blank_subcollections.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    # because we have a before create action that sets visibility to "open"
+    child.update_attributes(visibility: "restricted")
+  end
+
+  parent_collection.members << template
+  parent_collection.save!
+
+  # because we have a before create action that sets visibility to "open"
+  template.visibility = "restricted"
```
fenekku: So visibility can't be set at creation time?

lib/scripts/create_blank_subcollections.rb
```diff
@@ -0,0 +1,62 @@
+# Script for creating nested collections, for institutional collections.
+# Creates "x" 'template' collections, adding another "y" template
+# collections to each.
+# To use, pass in a collection id, a number for template collections amount,
+# and letter for child template collections amount. The 'letter' parameter is
+# used in the form of a range, starting from 'a'. i.e. if 'letter' is 'c',
+# range is a..c => a,b,c
+# Example to create 10 template collections with 3 collections for each:
+# current_sufia/ $ bundle exec rails runner lib/scripts/create_template_collections.rb "abc-123" 10 "c"
```
fenekku: Interesting choice to use a letter as input...  puts more work on the user though. And obviously if they forget and use a number, no subcollection will be created because numbers are before lowercase letters. Since you would be running the script it's fine.

sharpattack: Yea, this is my attempt to shorten the title for child collections. If you have a better idea, I am open to it. We can open an issue for it. 
Also, 
```suggestion
# current_sufia/ $ bundle exec rails runner lib/scripts/create_blank_subcollections.rb "abc-123" 10 "c"
```

lib/scripts/create_blank_subcollections.rb
sharpattack: Yes, it would be reverted to "open" if we try to set it before it was created.

sharpattack:
Sorry for merging before testing on staging and reporting, but the scripts worked without hiccups.

PR #730, opened 2020-03-02
(WIP) [#729] Use Different logout URL

### LINKED ISSUE #729: Fix typo in logout url
There was trouble with the url mentioned in #716. 

See XiaoXia's response after we switched to the #716 url and ended up at a page not found. 

> This application uses new SSO, not Shibboleth. Could you try to change the logout URL  to https://prd-nusso.it.northwestern.edu/nusso/XUI/?realm=/northwestern#logout/ 

After manually switching to the above url, logout seemed to be working as expected.

sharpattack:
The previous logout url did not work for our setup. closes #729

### DIFF
--- a/config/local_env_public.rb
+++ b/config/local_env_public.rb
@@ -10,7 +10,7 @@ ENV['LDAP_PORT']     = '636'
 
 begin
   if Rails.env == 'production'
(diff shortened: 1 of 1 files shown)


sharpattack:
Only after opening this PR did I catch my typo mistake. After testing the url to what it should have been, we will be using that one as it is the recommended one, and not the url in this PR.

PR #746, opened 2020-06-24
[#745] Add Publisher to 12th General Hospital

### LINKED ISSUE #745: Add Publisher to 12th General Hospital Items
Add "DigitalHub. Galter Health Sciences Library & Learning Center" to items in the 12th General Hospital collection with no publisher.

sharpattack:
Add default DigitalHub publisher to items in the 12th General Hospital
collection without a publisher. Create method that can be reused for
other collections. closes #745

### DIFF
--- a/lib/batch/add_publisher_for_collection_members.rb
+++ b/lib/batch/add_publisher_for_collection_members.rb
@@ -1,0 +1,15 @@
+# updates all the items in a collection with no publisher value.
+# Requires a collection_id
+
+def add_publisher_for_collection_members(collection_id:)
+  collection = Collection.find(collection_id)
+
+  puts "Checking #{collection.members.count} items in #{collection.title} for empty publisher"
+  collection.members.each do |member|
+    if member.class == Collection
+      add_publisher_for_collection_members(collection_id: member.id)
+    end
+
+    member.update(publisher: ["DigitalHub. Galter Health Sciences Library & Learning Center"]) if member.publisher.blank?
+  end
+end
--- a/lib/batch/add_publisher_to_12th_general_hospital_745.rb
+++ b/lib/batch/add_publisher_to_12th_general_hospital_745.rb
@@ -1,0 +1,5 @@
+require "#{Rails.root}/lib/batch/add_publisher_for_collection_members"
+# batch updates for https://github.com/galterlibrary/digital-repository/issues/745
+
+twelfth_general_hospital_id = "07b25bee-4a47-466a-b9b8-70d7a392fab0"
+add_publisher_for_collection_members(collection_id: twelfth_general_hospital_id)
--- a/spec/lib/batch/add_publisher_for_collection_members_spec.rb
+++ b/spec/lib/batch/add_publisher_for_collection_members_spec.rb
@@ -1,0 +1,53 @@
+require 'rails_helper'
+require "#{Rails.root}/lib/batch/add_publisher_for_collection_members"
+
+RSpec.describe 'rails runner lib/batch/add_publisher_for_collection_members.rb' do
+  let(:user) { FactoryGirl.create(:user) }
+  let(:item) {  make_generic_file(user) }
+  let(:collection) { make_collection(user) }
+  let(:nested_item) { make_generic_file(user) }
+  let(:nested_collection) { make_collection(user) }
+
+  before do
+    item.update_attributes(publisher: [])
+    collection.update_attributes(members: [item])
+    nested_item.update_attributes(publisher: [])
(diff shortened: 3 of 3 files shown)


sharpattack:
This took longer then it should have. I was banging my head at the wall for a while because I thought the tests were not working correctly. Turns out the tests were right, I just did not pay close attention to what I was trying to test.

lib/batch/add_publisher_for_collection_members.rb
```diff
@@ -0,0 +1,18 @@
+# updates all the items in a collection with no publisher value.
+# Requires a collection_id
+
+def add_publisher_for_collection_members(collection_id:)
+  collection = Collection.find(collection_id)
+
+  puts "Checking #{collection.members.count} items in #{collection.title} for empty publisher"
+  collection.members.each do |member|
+    if member.class == Collection
+      add_publisher_for_collection_members(collection_id: member.id)
```
Meowcenary, line 10: Interesting recursion here. Just so I'm clear , when `member` is a `Collection` is that representing when a `member` owns a group of things?

lib/batch/add_publisher_for_collection_members.rb
```diff
@@ -0,0 +1,18 @@
...
+    end
+
+    if member.publisher.empty?
+      member.publisher += ["DigitalHub. Galter Health Sciences Library & Learning Center"]
+      member.save!
```
Meowcenary, line 15: Two comments about this, but I don't feel strongly about either. Just things to consider.

I'd recommend using `#blank?` and not `#empty?` here. The difference of these is that empty will only return true if it's a completely empty string or nil, but blank will also return true if it's all white space characters. This will probably not matter here, but it's something to consider.

It's not really a big difference, but you can save a line by doing `member.update!(publisher: "...")`. You could save even more lines with a post conditional: `member.update(publisher: "") if member.publisher.blank?`.

lib/batch/add_publisher_to_12th_general_hospital_745.rb
```diff
@@ -0,0 +1,5 @@
+require "#{Rails.root}/lib/batch/add_publisher_for_collection_members"
```
Meowcenary, line 1: I've never seen this before. Pretty slick way to use relative paths.

spec/lib/batch/add_publisher_for_collection_members_spec.rb
```diff
@@ -0,0 +1,52 @@
+require 'rails_helper'
+require "#{Rails.root}/lib/batch/add_publisher_for_collection_members"
+
+RSpec.describe 'rails runner lib/batch/add_publisher_for_collection_members.rb' do
+  before do
+    @user = FactoryGirl.create(:user)
```
Meowcenary, line 6: I don't expect you to re-write anything here, but we should absolutely move away from using member variables (I think that's what they call it when you use @) and instead try to use `let` blocks where we can. It's a pretty standard practice in newer rails apps and there are a few reasons for it that I won't get into here.

lib/batch/add_publisher_for_collection_members.rb
Meowcenary: Ah, after reading the spec I think this is meant to handle nested Collections. Seemingly that's when a Collection has a Collection within it right?

sharpattack: Yes, exactly that. It's possible Collections can have other Collections, in addition to GenericFiles.

Meowcenary:
> This took longer then it should have. I was banging my head at the wall for a while because I thought the tests were not working correctly. Turns out the tests were right, I just did not pay close attention to what I was trying to test.

I missed that on my first run through of this, but saw it while clearing out notifications. Just wanted to say glad you figured it out. I've had that happen a lot recently too ha.

lib/batch/add_publisher_for_collection_members.rb
sharpattack: Ah, I do not think I ever knew the difference with `#blank?` and `#empty?`. I like the update suggestion, so I will make the change.

spec/lib/batch/add_publisher_for_collection_members_spec.rb
sharpattack: Nice to know, I did not realize that. Please do continue to suggest changes, it is good to be aware of these things.

lib/batch/add_publisher_for_collection_members.rb
Meowcenary: Got it. Yeah I think that's a testament to the quality of your tests and code that I was able to deduce that. Good stuff Austin.

PR #774, opened 2020-09-03
[#771] Export Collections Tags and Mesh

### LINKED ISSUE #771: Export Collections Keywords and Subject: MeSH
Request for a CSV export of DigitalHub Collections and all sub-collections' Keywords and Subject: MeSH terms/

sharpattack:
Export collections tags and mesh info to csv file using rake.
closes #771

### DIFF
--- a/lib/scripts/collections_tags_and_mesh.rb
+++ b/lib/scripts/collections_tags_and_mesh.rb
@@ -1,0 +1,39 @@
+require 'csv'
+
+class CollectionsTagsAndMeshList
+  attr_accessor :csv_file
+
+  def initialize
+    @csv_file = File.new("#{Rails.root}/lib/scripts/results/collections_tags_and_mesh_list.csv", "w+")
+  end
+
+  def populate_collections_tags_and_mesh_info
+    CSV.open(self.csv_file.path, "w") do |csv|
+      # add headers
+      csv << ['Title', 'Link', 'Tags', 'MeSH']
+
+      select_all_collections.each do |collection|
+        # remove last "_" part of string
+        collection_data = collection.each.map {|k, v| [k.gsub(/_[a-z]+\z/, ''), v] }.to_h
+        collection_url = "https://digitalhub.northwestern.edu/collections/#{collection_data['id']}"
+
+        csv << [clean_up_data(collection_data["title"]),
+                collection_url,
+                clean_up_data(collection_data["tag"]),
+                clean_up_data(collection_data["mesh"])]
+      end
+    end # end block, CSV closes
+  end
+
+  private
+  def select_all_collections
+    # Collection is a fedora object, so using ActiveFedora to get all
+    # Collections is slow, and takes up too much memory. Querying solr is much
+    # faster, takes up less memory, and still includes the info we want.
+    ActiveFedora::SolrService.query('has_model_ssim:Collection', { rows: 99999 } )
+  end
+
+  def clean_up_data(value)
+    (value.is_a?(Array) ? value.reject(&:blank?).compact.join(" ; ") : value).to_s
+  end
+end
--- a/spec/lib/scripts/collections_tags_and_mesh_spec.rb
+++ b/spec/lib/scripts/collections_tags_and_mesh_spec.rb
@@ -1,0 +1,33 @@
+require 'rails_helper'
+require "#{Rails.root}/lib/scripts/collections_tags_and_mesh"
+
+RSpec.describe 'lib/scripts/collections_tags_and_mesh.rb' do
+  let(:user) { FactoryGirl.create(:user) }
+
+  subject { CollectionsTagsAndMeshList.new }
+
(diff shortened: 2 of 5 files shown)


lib/scripts/collections_tags_and_mesh.rb
```diff
@@ -0,0 +1,38 @@
...
+                "https://digitalhub.northwestern.edu/collections/#{collection_data['id']}",
+                clean_up_data(collection_data, "tag"),
+                clean_up_data(collection_data, "mesh")]
+      end
+    end # end block, CSV closes
+  end
+
+  private
+  def select_all_collections
+    # Collection is a fedora object, so using ActiveFedora to get all
```
Meowcenary, line 29: Good to know. Thanks for writing this out.

lib/scripts/collections_tags_and_mesh.rb
```diff
@@ -0,0 +1,38 @@
...
+    CSV.open(self.csv_file.path, "w") do |csv|
+      # add headers
+      csv << ['Title', 'Link', 'Tags', 'MeSH']
+
+      select_all_collections.each do |collection|
+        # remove last "_" part of string
+        collection_data = collection.each.map {|k, v| [k.gsub(/_[a-z]+\z/, ''), v] }.to_h
+
+        csv << [clean_up_data(collection_data, "title"),
+                "https://digitalhub.northwestern.edu/collections/#{collection_data['id']}",
```
Meowcenary, line 20: It might make sense to put this into a variable naming the concept e.g `digital_hub_url = "https://digitalhub.northwestern.edu/collections/#{collection_data['id']}"` and then `csv << [..., digital_hub_url, ...]` . Not a big deal either way.

lib/scripts/collections_tags_and_mesh.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  private
+  def select_all_collections
+    # Collection is a fedora object, so using ActiveFedora to get all
+    # Collections is slow, and takes up too much memory. Querying solr is much
+    # faster, takes up less memory, and still includes the info we want.
+    ActiveFedora::SolrService.query('has_model_ssim:Collection', { rows: 99999 } )
+  end
+
+  def clean_up_data(data, key)
```
Meowcenary, line 35: instead of passing data and key as separate arguments and then calling into data[key] several times in the function I would change this to something like

```
def clean_up_data(value)
  (value.is_a?(Array) ? value.reject(&:blank?).compact.join(" ; ") : data[key]).to_s
end
```

Then call it with `clean_up_data(collection_data["mesh"])`

spec/lib/scripts/collections_tags_and_mesh_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+                        "https://digitalhub.northwestern.edu/collections/123,"\
+                        "CERN,invenioRDM\n"
+
+        expect(collections_tags_and_mesh_list_result[1]).to eq(expected_data)
+      end
+    end # #populate_user_info
+  end
+
+  after do
+    FileUtils.rm_rf(Dir["#{Rails.root}/lib/scripts/results/collections_tags_and_mesh_list.csv"])
```
Meowcenary, line 31: I would use [`FileUtils.rm_f`](https://ruby-doc.org/stdlib-2.4.1/libdoc/fileutils/rdoc/FileUtils.html#method-c-rm_f) since it's for single files instead of `FileUtils.rm_rf` which I think is for directories and contents. I'm not sure what the vulnerability mentioned in `FileUtils.rm_rf`'s documentation is, but might be worth considering too.

Meowcenary (approved):
Overall looks good. Two small comments for you to consider.

spec/lib/scripts/collections_tags_and_mesh_spec.rb
sharpattack: Ah, good idea. I didn't realize the vulnerability, but this is better anyways.

lib/scripts/collections_tags_and_mesh.rb
sharpattack: Much better. To only pass the specific data, instead of the entire hash each time.

lib/scripts/collections_tags_and_mesh.rb
sharpattack: Sure, I can make this change.

spec/lib/scripts/collections_tags_and_mesh_spec.rb
Meowcenary: To be honest with you, I didn't even read _what_ the vulnerability was lol. I just figured that it was better to 1: avoid whatever the vulnerability was and 2: not leave room for us to potentially remove an entire directory by accident.

Meowcenary (approved):
Thank for taking care of this initially and updating it. Looks good to me.

PR #777, opened 2020-11-11
[#775] CSV of Collection Items

### LINKED ISSUE #775: Get CSV list of items in a Collection
With Collection title, the item title, and the url to the item.

sharpattack:
Export a list of items in a collection. closes #775

### DIFF
--- a/spec/lib/scripts/collection_items_list_spec.rb
+++ b/spec/lib/scripts/collection_items_list_spec.rb
@@ -1,0 +1,82 @@
+require 'rails_helper'
+require "#{Rails.root}/lib/scripts/collection_items_list"
+
+RSpec.describe 'lib/scripts/collection_items_list.rb' do
+  let(:user) { FactoryGirl.create(:user) }
+  let(:file) {
+    make_generic_file(user, { id: 'bogus-file-id', title: ['Test File'] })
+  }
+  let(:parent_collection) {
+    make_collection(user, { title: 'Test Collection'})
+  }
+
+  before do
+    parent_collection.members << file
+    parent_collection.save!
+  end
+
+  describe 'CollectionItemsList' do
+    subject { CollectionItemsList.new(parent_collection.id) }
+
+    it '#initialize' do
+      expect(subject.collection.present?).to be_truthy
+      expect(subject.collection_items_csv_file.present?).to be_truthy
+    end
+
+    it 'creates the file with sanitized title' do
+      expect(
+        File.basename(subject.collection_items_csv_file.path)
+      ).to eq("Test_Collection_items_list.csv")
+    end
+
+    describe '#get_items_and_add_to_csv' do
+      context 'collection with one file' do
+        it 'adds items data to Test_Collection_items_list.csv' do
+          subject.get_items_and_add_to_csv
+          csv_result = File.readlines(
+            "#{Rails.root}/lib/scripts/results/Test_Collection_items_list.csv"
+          )
+          expected_data = "#{parent_collection.title},"\
+                          "#{file.title.first},"\
+                          "https://digitalhub.northwestern.edu/files/#{file.id}\n"
+          expect(csv_result[1]).to eq(expected_data)
+        end
+      end
+
+      context 'collection with one file and one child collection with one file' do
+        let(:child_file) {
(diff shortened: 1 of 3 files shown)


spec/lib/scripts/collection_items_list_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+  before do
+    parent_collection.members << file
+    parent_collection.save!
+  end
+
+  describe 'CollectionItemsList' do
+    subject { CollectionItemsList.new(parent_collection.id) }
+
+    it '#initialize' do
+      expect(subject.collection).to be_truthy
```
Meowcenary, line 22: Is this just checking presence? If that's the case it might make more sense to use `expect(subject.collection.present?).to be_truthy` or something along those lines. This would just to make it clear that it's a test for presence only. This is very nit picky since most people coding in ruby would know the truthy values anyways so I'll leave it up to you how you want to handle this.

spec/lib/scripts/collection_items_list_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      context 'collection with one file and one child collection with one file' do
+        let(:child_file) {
+          make_generic_file(user, { id: 'bogus-child-file-id', title: ['Test Child File'] })
+        }
+        let(:child_collection) {
+          make_collection(user, {title: 'Test Child Collection'})
+        }
+
+        before do
+          child_collection.members << child_file
```
Meowcenary, line 55: I think you can one line these as something like `child_collection.update!(members: [child_file])`

spec/lib/scripts/collection_items_list_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+          child_collection.save!
+          parent_collection.members << child_collection
+          parent_collection.save!
+        end
+
+        it 'adds items data to Test_Collection_items_list.csv' do
+          subject.get_items_and_add_to_csv
+          csv_result = File.readlines(
+            "#{Rails.root}/lib/scripts/results/Test_Collection_items_list.csv"
```
Meowcenary, line 64: It might make sense to use a fixture here rather than dynamically generating the test collection each time. I'm not sure how much time this would save, probably only a few seconds, so again I'll leave this up to you.

Meowcenary (approved):
Thanks for handling this. Good idea making a class to help with this in the future too. Left a few comments, but nothing that should stop you from merging.

spec/lib/scripts/collection_items_list_spec.rb
sharpattack: I see what you mean. I think I was just testing that you can call `#collection` and `#collection_items_csv_file` on the object. It is not as clear that's what I am really doing, but checking for the presence will be.

spec/lib/scripts/collection_items_list_spec.rb
sharpattack: I believe I had trouble trying it this way in the past due to this being an ActiveFedora object, and not ActiveRecord. It would be better to have it in one line, so I will try again.

spec/lib/scripts/collection_items_list_spec.rb
sharpattack: Did you mean make a fixture for the csv file? I think that's what I want to test. When calling `#get_items_and_add_to_csv` it creates the file with results. Unless you mean to make something else a fixture?

spec/lib/scripts/collection_items_list_spec.rb
Meowcenary: Oh, I didn't realize it was not ActiveRecord. I'm not familiar with ActiveFedora so I will leave it to you to decide the best course of action here.

spec/lib/scripts/collection_items_list_spec.rb
Meowcenary: Ah, no, I get what you mean. I think I was looking at this test the wrong way.

carsonicator (approved):
Thanks, Austin.

spec/lib/scripts/collection_items_list_spec.rb
sharpattack: Yea, this didn't work 😞

PR #789, opened 2020-12-16
[#788] Create Sufia Export Subclass InvenioRdmConverter

### LINKED ISSUE #788: Create Sufia Export Subclass InvenioRdmConverter
Sufia 6.x (the branch that is) and version 6.7 have classes and an executable that can be used to export a JSON of the repo's data.  In order to override the default behavior of this class for custom data models you must define a subclass from the gem defined "converter" class. Create this subclass and wrap the executable call into a rake task.

Meowcenary:
Create skeleton subclass from Sufia export converter to handle
InvenioRDM specific export. Create a rake task to encapsulate parameters
specific to this export. closes #788

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -417,7 +417,7 @@ GEM
       sparql (~> 1.99)
       sparql-client (~> 1.99)
     little-plugger (1.1.4)
-    logger (1.2.8)
+    logger (1.2.8.1)
     logging (2.2.2)
       little-plugger (~> 1.1)
       multi_json (~> 1.10)
--- a/app/models/invenio_rdm_collection_converter.rb
+++ b/app/models/invenio_rdm_collection_converter.rb
@@ -1,0 +1,18 @@
+include Sufia::Export
+
+# Convert a Collection including metadata, permissions and version metadata into a PORO
+# so that the metadata can be exported in json format using to_json
+#
+class InvenioRdmCollectionConverter < Sufia::Export::Converter
+  # Create an instance of a InvenioRdmCollectionConverter converter containing all the metadata for json export
+  #
+  # @param [collection] collection file to be converted for export
+  def initialize(collection)
+    @id = collection.id
+    # @title = collection.title
+    # @description = collection.description
+    # @creator = collection.creator.map { |c| c }
+    # @members = collection.members.map(&:id)
+    # @permissions = permissions(collection)
+  end
+end
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -1,0 +1,48 @@
+include Sufia::Export
+
+# Convert a GenericFile including metadata, permissions and version metadata into a PORO
+# so that the metadata can be exported in json format using to_json
+#
+class InvenioRdmRecordConverter < Sufia::Export::Converter
+  # Create an instance of a InvenioRdmRecordConverter converter containing all the metadata for json export
+  #
+  # @param [GenericFile] generic_file file to be converted for export
+  def initialize(generic_file)
+    @id = generic_file.id
+    @provenance = generic_file.depositor
+    # @label = generic_file.label
+    # @depositor = generic_file.depositor
+    # @arkivo_checksum = generic_file.arkivo_checksum
+    # @relative_path = generic_file.relative_path
+    # @import_url = generic_file.import_url
+    # @resource_type = generic_file.resource_type
+    # @title = generic_file.title
+    # @creator = generic_file.creator
+    # @contributor = generic_file.contributor
+    # @description = generic_file.description
+    # @tag = generic_file.tag
+    # @rights = generic_file.rights
+    # @publisher = generic_file.publisher
+    # @date_created = generic_file.date_created
+    # @date_uploaded = generic_file.date_uploaded
+    # @date_modified = generic_file.date_modified
+    # @subject = generic_file.subject
+    # @language = generic_file.language
+    # @identifier = generic_file.identifier
+    # @based_near = generic_file.based_near
+    # @related_url = generic_file.related_url
+    # @bibliographic_citation = generic_file.bibliographic_citation
+    # @source = generic_file.source
+    # @batch_id = generic_file.batch.id if generic_file.batch
+    # @visibility = generic_file.visibility
+    # @versions = versions(generic_file)
+    # @permissions = permissions(generic_file)
+  end
+
+  private
+
+    def versions(gf)
+      return [] unless gf.content.has_versions?
+      Sufia::Export::VersionGraphConverter.new(gf.content.versions).versions
+    end
+end
--- a/lib/tasks/repo_export.rake
+++ b/lib/tasks/repo_export.rake
@@ -1,0 +1,9 @@
+desc "Run sufia export with InvenioRDM specific converters"
+task :repo_export, [:destination] => :environment do |task, args|
+  export_command = "sufia_export --verbose --models GenericFile=InvenioRdmRecordConverter,Collection=InvenioRdmCollectionConverter"
+  system export_command
+
+  if destination = args[:destination]
+    system "mv tmp/export/ #{destination}"
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -113,6 +113,15 @@ ActiveRecord::Schema.define(version: 20200813221318) do
   add_index "follows", ["followable_id", "followable_type"], name: "fk_followables", using: :btree
   add_index "follows", ["follower_id", "follower_type"], name: "fk_follows", using: :btree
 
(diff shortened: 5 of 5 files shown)


Meowcenary:
After much struggle the wheels of progress are turning!

Gemfile.lock
```diff
@@ -417,7 +417,7 @@ GEM
       sparql (~> 1.99)
       sparql-client (~> 1.99)
     little-plugger (1.1.4)
-    logger (1.2.8)
+    logger (1.2.8.1)
```
Meowcenary, line 420: Seems that logger 1.2.8 was removed from Rubygems, but this version does exist. If you already had logger 1.2.8 installed you would be fine bundling, but I did not have this version for whatever reason.

app/models/invenio_rdm_converter.rb
```diff
@@ -0,0 +1,48 @@
...
+# Convert a GenericFile including metadata, permissions and version metadata into a PORO
+# so that the metadata can be exported in json format using to_json
+#
+class InvenioRdmConverter < Sufia::Export::Converter
+  # Create an instance of a GenericFile converter containing all the metadata for json export
+  #
+  # @param [GenericFile] generc_file file to be converted for export
+  def initialize(generc_file)
+    @id = generc_file.id
+    @provenance = generc_file.depositor
```
Meowcenary, line 12: This is obviously not correct, but I think it's a good example of how you could set a custom field for the export. I'm leaving the commented out fields so we can see what the default export was.

lib/tasks/repo_export.rake
```diff
@@ -0,0 +1,9 @@
+desc "Run sufia export with Galter specific parameters"
```
Meowcenary, line 1: Just a convenience thing. I think it actually runs slow like this, but at least it stores the command we expect to run outside of our own heads.

app/models/invenio_rdm_converter.rb
sharpattack: Thanks for pointing this out.

app/models/invenio_rdm_converter.rb
```diff
@@ -0,0 +1,48 @@
+include Sufia::Export
+
+# Convert a GenericFile including metadata, permissions and version metadata into a PORO
+# so that the metadata can be exported in json format using to_json
+#
+class InvenioRdmConverter < Sufia::Export::Converter
```
sharpattack, line 6: So instead of using "GenericFile", we use this for invenioRDM record? Could we rename this to `InvenioRdmRecordConverter` maybe? Then do the same for "Collection", instead name the class `InvenioRdmCommunityConverter`.

lib/tasks/repo_export.rake
sharpattack: The command `sufia_export` loads the rails environment. Just curious if this could be an issue for the slowness? 

I think it would also help to be more specific here by adding "invenio" to the description.

app/models/invenio_rdm_converter.rb
```diff
@@ -0,0 +1,48 @@
...
+  # Create an instance of a GenericFile converter containing all the metadata for json export
+  #
+  # @param [GenericFile] generc_file file to be converted for export
+  def initialize(generc_file)
```
sharpattack, line 10:
```suggestion
  def initialize(generic_file)
```

sharpattack (commented):
Thanks for adding, looking forward to making it invenioRDM specific! Just a couple questions before approving.

app/models/invenio_rdm_converter.rb
Meowcenary: We can renname it , that makes sense to me. It's still technically using the GenericFile class from ActiveFedora (sufia?), but it's just storing it in a PORO that has different labels and functions that we want to use to make the conversions.

app/models/invenio_rdm_converter.rb
Meowcenary: yeah I was thinking of making this change too, but I just went with what I copied directly from the samvera/sufia repo. I think since you're on board I will change it.

lib/tasks/repo_export.rake
Meowcenary: Ah, that might be it. I was thinking it might have to do with having rake add extra steps for making the call, particularly the `system` function from ruby. Honestly I don't think there is any way that this export can be fast so I'm not terribly concerned with it.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -0,0 +1,48 @@
...
+# Convert a GenericFile including metadata, permissions and version metadata into a PORO
+# so that the metadata can be exported in json format using to_json
+#
+class InvenioRdmRecordConverter < Sufia::Export::Converter
+  # Create an instance of a InvenioRdmRecordConverter converter containing all the metadata for json export
+  #
+  # @param [GenericFile] generic_file file to be converted for export
+  def initialize(generic_file)
+    @id = generc_file.id
+    @provenance = generc_file.depositor
```
sharpattack, line 12: Sorry, I shouldn't have said anything, but all these `generc_file` need to change, too. 😞

app/models/invenio_rdm_collection_converter.rb
```diff
@@ -0,0 +1,18 @@
...
+  # @param [collection] collection file to be converted for export
+  def initialize(collection)
+    @id = collection.id
+    # @title = collection.title
+    # @description = collection.description
+    # @creator = collection.creator.map { |c| c }
+    # @members = collection.members.map(&:id)
+    # @permissions = permissions(collection)
+  end
+end
```
sharpattack, line 18: Thank you for adding!

lib/tasks/repo_export.rake
```diff
@@ -0,0 +1,9 @@
+desc "Run sufia export with InvenioRDM specific converters"
+task :repo_export, [:destination] => :environment do |task, args|
+  export_command = "sufia_export --verbose --models GenericFile=InvenioRdmConverter,Collection=InvenioRdmCollectionConverter"
```
sharpattack, line 3: Should the option be `GenericFile=InvenioRdmRecordConverter`?
```suggestion
  export_command = "sufia_export --verbose --models GenericFile=InvenioRdmRecordConverter,Collection=InvenioRdmCollectionConverter"
```

sharpattack (commented):
Thanks for making the changes! But, there are a couple small things left.

app/models/invenio_rdm_collection_converter.rb
```diff
@@ -0,0 +1,18 @@
+include Sufia::Export
+
+# Convert a Collection including metadata, permissions and version metadata into a PORO
+# so that the metadata can be exported in json format using to_json
+#
+class InvenioRdmCollectionConverter < Sufia::Export::Converter
+  # Create an instance of a InvenioRdmCollectionConverter converter containing all the metadata for json export
+  #
+  # @param [collection] collection file to be converted for export
+  def initialize(collection)
```
carsonicator, line 10: Just curious, are the variables in the 'initialize' function commented out because you're testing?

app/models/invenio_rdm_record_converter.rb
```diff
@@ -0,0 +1,48 @@
+include Sufia::Export
+
+# Convert a GenericFile including metadata, permissions and version metadata into a PORO
+# so that the metadata can be exported in json format using to_json
+#
+class InvenioRdmRecordConverter < Sufia::Export::Converter
+  # Create an instance of a InvenioRdmRecordConverter converter containing all the metadata for json export
+  #
+  # @param [GenericFile] generic_file file to be converted for export
+  def initialize(generic_file)
```
carsonicator, line 10: same here about 'initialize'?

carsonicator (approved):

lib/tasks/repo_export.rake
Meowcenary: Ah, it should be. Thanks.

app/models/invenio_rdm_record_converter.rb
Meowcenary: Oh god, I really rushed this. Well, it's a good thing you double checked.

app/models/invenio_rdm_collection_converter.rb
Meowcenary: I'm leaving the commented out variables so we have an easy reference to the fields that are available from ActiveFedora  on the Collection and GenericFile objects. For me personally I like to have it in the code so I don't have to open a console or a webpage.

app/models/invenio_rdm_record_converter.rb
Meowcenary: Same as above. Just trying to keep thins where it will be easy to reference.

Meowcenary:
@sharpattack , let me know when you're good on this and I'll merge.

sharpattack (approved):
Looks good, thanks for making the review suggested changes.

Issue #785, opened 2020-11-16
Export Invenio Provenance Field


Meowcenary:
Export the [top level field](https://inveniordm.docs.cern.ch/reference/metadata/#overview) `provenance` from the Sufia format to the Invenio format. Use the following as test data to verify the end result:

```
        "provenance": {
            "created_by": {
                "user": 1
            },
            "on_behalf_of": {
                "user": 2
            }
        }
```

PR #807, opened 2021-01-07
[#782] Export PID Field. closes #782

### LINKED ISSUE #782: Export Invenio Pids Field
Export the [top level field](https://inveniordm.docs.cern.ch/reference/metadata/#overview) `pids` from the Sufia format to the Invenio format. Use the following as test data to verify the end result:

```
        "pids": {
            "doi": {
                "identifier": "10.5281/zenodo.1234",
                "provider": "datacite",
                "client": "zenodo"
            },
            "concept-doi": {
                "identifier": "10.5281/zenodo.1234",
                "provider": "datacite",
                "client": "zenodo"
            },
            "handle": {
                "identifier": "9.12314",
                "provider": "cern-handle",
                "client": "zenodo"
            },
            "oai": {
                "identifier": "oai:zenodo.org:12345",
                "provider": "zenodo"
            }
        }
```

sharpattack:


sharpattack:
Went with a much more simplified approach for testing.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'rails_helper'
+
+RSpec.describe InvenioRdmRecordConverter do
+  let(:user) { FactoryGirl.create(:user) }
+  let(:generic_file) { make_generic_file(user, doi: ["doi:123/ABC"]) }
+  let(:json) { "{\"pids\":{\"doi\":{\"identifier\":\"#{generic_file.doi.shift}\",\"provider\":\"datacite\",\"client\":\"digitalhub\"}},\"provenance\":\"#{user.username}\"}" }
```
sharpattack, line 6: As we add new fields to `InvenioRdmRecordConverter`, we can just add our expected values to this.

Meowcenary: Sounds good to me. Is provenance marked with # because we're not sure what to put in this field currently? I'm going to follow your lead here after you merge to make my own specs and I think that would be a decent standard for fields we don't know how to deal with right now. Similarly, for ambiguous mappings I think we should have some sort of standard for how we delimit things. Right now i'm using semi-colons to separate, but I think i'll switch it to commas so it's more in line with csv. You will see what i'm talking about later when I open that PR, but it's kind of an odd situation.

sharpattack: Oh, i just added provenance here as it was in `InvenioRdmRecordConverter`. I did not realize it was not what we want, but I can address that in #785

Meowcenary (approved):
Whoops, I forgot to approve.

spec/models/invenio_rdm_record_converter_spec.rb
Meowcenary: Oh, I see. I sort of forgot what was there, but yeah I think for now we can leave the # and just address it when the time comes.

PR #808, opened 2021-01-07
[#1790] Export Metadata Subfield Resource Type

Meowcenary:
closes #1790


app/models/concerns/galtersufia/generic_file/invenio_resource_type_mappings.rb
```diff
@@ -0,0 +1,180 @@
...
+        "Tables": "Table",
+        "Technical Report": "Technical Documentation",
+      }.with_indifferent_access
+
+      # Mapping of IRDM sub type / specific resource type to IRDM resource type / category
+      IRDM_RESOURCE_TYPES = {
+        "Collection": "Archival Items",
+        "Ephemera": "Archival Items",
+        "Exhibition": "Archival Items",
+        # "Physical Object": "Archival Items",
```
Meowcenary, line 87: Commented out fields are invenio specific types -> invenio types that did not match any DH values or records. In other words unless someone were to add something to DH that would then map to one of these commented out fields then we would have a single record using it. I figure it's not a bad idea to keep them here since I don't want to retype this again (after accidentally screwing something up with it and having to audit it x_x). If it's really causing problems we can remove these lines.

app/models/concerns/galtersufia/generic_file/invenio_resource_type_mappings.rb
```diff
@@ -0,0 +1,180 @@
...
+        "Research Paper": "Report; Review; Other Article",
+        "Resource Guides": "Resource Guide",
+        "Review": "Review",
+        "Software or Program Code ": "Other; Software; Technical Documentation",
+        "Speech": "Lecture; Presentation; Sound (Recording)",
+        "Statistics": "Statistics",
+        "Study Design": "Study Design",
+        "Tables": "Table",
+        "Technical Report": "Technical Documentation",
+      }.with_indifferent_access
```
Meowcenary, line 80: [HashWithIndifferentAccess](https://api.rubyonrails.org/classes/ActiveSupport/HashWithIndifferentAccess.html) is a fancily named rails class that is really just a hash you can access with symbols OR with strings for keys. Normally hashes only support symbols because they are immutable, but I think for our use case it's fine to do it this way. If nothing else I'm not really sure what the alternative is.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -2,8 +2,25 @@
...
+        "doi": {
+          "identifier": "#{generic_file.doi.shift}",
+          "provider":"datacite",
+          "client":"digitalhub"
+        }
+      },
+      "provenance":"#{user.username}",
+      "metadata": {
+        "resource_type": {
+          "type": "Books",
```
Meowcenary, line 18: There are types that can be assigned for a file in DH that are not being used at all. For example, textbooks are supported, but not used. For now I think this is a good enough test, but at some point it would probably be good to either sample an array of types that we know we need to support OR modify the factory to create more generic files to convert. I think the latter idea is the better one, but will take more effort. Testing stuff always causes me problems it seems.

carsonicator (approved):

app/models/concerns/galtersufia/generic_file/invenio_resource_type_mappings.rb
sharpattack: Nice! I did not know that about HashWithIndifferentAccess. I like it, and I hope I can remember to use it for a future case. It looks good to me here.

app/models/concerns/galtersufia/generic_file/invenio_resource_type_mappings.rb
sharpattack: Thanks for adding all of this! I appreciate the work you did here. Good idea adding and commenting out all of the other fields that are only in invenio.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -50,4 +53,20 @@ def versions(gf)
       return [] unless gf.content.has_versions?
       Sufia::Export::VersionGraphConverter.new(gf.content.versions).versions
     end
+
+    def invenio_metadata(gf)
```
sharpattack, line 57: I was going to say just remove this method, and have the following three lines on line 20, but I think this looks much cleaner.

spec/models/invenio_rdm_record_converter_spec.rb
sharpattack: Thanks for the explanation. I think this is good enough for now, but let's make a note to improve tests down the road if we are able. And thanks for redefining `json` here, I don't know what I was thinking keeping it one-line.

sharpattack (approved):
Great work, I appreciate the changes to the test.

PR #810, opened 2021-01-11
[#809] Revert Resque in Development

### LINKED ISSUE #809: Revert Resque to 1.26
When trying to run Resque 1.27.4, we receive the following error:
```
rake aborted!
ActiveSupport::Concern::MultipleIncludedBlocks: Cannot define multiple 'included' blocks for a Concern
```

In __development and testing__ only, revert to previous working version. The upgraded version of Resque that includes a fix for eager loading does not support rails 4 or ruby 2.3. https://github.com/resque/resque/releases/tag/v2.0.0

Placing the Resque version decrease in the development group should be enough, as in production and staging, `eager_load` is set to true.

sharpattack:
In development and testing, eager_load is set to false which causes
errors when trying to run resque-1.27.4. Staging and Production has
eager_load set to true. closes #809

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -67,6 +67,7 @@ group :development, :test do
   gem 'flamegraph'
   gem 'memory_profiler'
   gem 'timecop'
+  gem 'resque', '1.26.0'
 end
 
 gem "better_errors",  group: :development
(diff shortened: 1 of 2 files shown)


Meowcenary (approved):
Looks good to me. Thanks for fixing this.

PR #812, opened 2021-01-11
[#791] Export Metadata Subfield Creators

### LINKED ISSUE #791: Export Metadata Subfield Creators
Export the Metadata subfield [Creators](https://inveniordm.docs.cern.ch/reference/metadata/#creators-1-n)

Meowcenary:
closes #791

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -5,11 +5,14 @@ include Sufia::Export
 #
 class InvenioRdmRecordConverter < Sufia::Export::Converter
   include Galtersufia::GenericFile::InvenioResourceTypeMappings
+  include Galtersufia::GenericFile::KnownOrganizations
 
   # Create an instance of a InvenioRdmRecordConverter converter containing all the metadata for json export
   #
   # @param [GenericFile] generic_file file to be converted for export
-  def initialize(generic_file)
+  def initialize(generic_file=nil)
+    return unless generic_file
+
     #PIDs
     @pids = invennio_pids(generic_file.doi.shift)
     @metadata = invenio_metadata(generic_file)
@@ -60,22 +63,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
     }
   end
 
-  def invenio_metadata(gf)
-    {
-      "resource_type": resource_type(gf)
-    }
-  end
-
-  def resource_type(gf)
-    irdm_subtype = DH_RESOURCE_TYPES[gf.resource_type.shift]
-    irdm_type = IRDM_RESOURCE_TYPES[irdm_subtype]
-
-    {
-      "type": irdm_type,
-      "subtype": irdm_subtype
-    }
-  end
-
   def invenio_provenance(proxy_depositor, on_behalf_of)
     {
       "created_by": {
@@ -85,5 +72,72 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
         "user": on_behalf_of
       }
     }
+  end
+
+  def invenio_metadata(gf)
+    {
+      "resource_type": resource_type(gf.resource_type.shift),
+      "creators": creators(gf.creator)
+    }
+  end
+
+
+  def resource_type(digitalhub_subtype)
+    irdm_subtype = DH_RESOURCE_TYPES[digitalhub_subtype]
+    irdm_type = IRDM_RESOURCE_TYPES[irdm_subtype]
+
+    {
+      "type": irdm_type,
+      "subtype": irdm_subtype
+    }
+  end
+
+  def creators(creators)
+    creators.map do |creator|
+      # Galter Health Sciences Library is a User on production DigitalHub, but it does NOT have a formal_name value
+      dh_user = User.find_by(formal_name: creator)
+
+      # organization
+      if organization?(creator)
+        name = creator
+        creator_type = "organisational"
+        family_name = ""
+        given_name = ""
+        identifiers = {}
+        affiliations = [] # only used for creator_type "personal"
+      # user
+      elsif dh_user.present?
+        name = dh_user.formal_name # will always be this regardless of type
+        creator_type = "personal"
+        dh_user_display_name = dh_user.display_name.split(' ') # split name into components to be reused
+        family_name = dh_user_display_name.pop # remove last value from display name
+        given_name = dh_user_display_name.join(' ') # the remaining strings becomes given name
+        identifiers = dh_user.orcid.present? ? {"orcid": dh_user.orcid.split('/').pop} : {}
+        affiliations = [] # TODO: user.affiliation is nil for all users on production... where do we get this value?
+      # TODO: unknown personal, what do we enter for these fields?
+      else
+        name = creator
+        creator_type = "personal"
+        family_name = ""
+        given_name = ""
+        identifiers = {}
+        affiliations = [] # TODO: user.affiliation is nil for all users on production... where do we get this value?
+      end
+
+      {
+        "name": name.include?("not identified") ? "" : creator,
+        "type": creator_type,
+        "role": "",
+        "given_name": given_name,
+        "family_name": family_name,
+        "identifiers": identifiers, # TODO: where can we find additional identifiers?
+        "affiliations": affiliations
+      }
+    end
+  end
+
+  def format_creator_names(creator_name)
+    family_name = creator_name.pop # remove last value from display name
+    given_name = creator_name.join(' ') # the remaining strings becomes given name
   end
 end
--- a/spec/models/invenio_rdm_record_converter_spec.rb
+++ b/spec/models/invenio_rdm_record_converter_spec.rb
@@ -1,7 +1,8 @@
 require 'rails_helper'
 
 RSpec.describe InvenioRdmRecordConverter do
-  let(:user) { FactoryGirl.create(:user, username: "usr1234") }
+  let(:user) { FactoryGirl.create(:user, username: "usr1234", formal_name: "Tester, Mock", orcid: "https://orcid.org/1234-5678-9123-4567",
+                                    display_name: 'Mock Tester') }
   let(:assistant) { FactoryGirl.create(:user, username: "ast9876") }
 
   before do
@@ -14,9 +15,11 @@ RSpec.describe InvenioRdmRecordConverter do
       doi: ["doi:123/ABC"],
       resource_type: ["Account Books"],
       proxy_depositor: assistant.username,
-      on_behalf_of: user.username
+      on_behalf_of: user.username,
+      creator: [user.formal_name]
     )
   }
+
   let(:json) do
     {
       "pids": {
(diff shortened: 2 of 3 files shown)


Meowcenary (commented):
There are a handful of questions under the TODO comments that I'm going to bring up to Sara and Guillaume. To recap them here:

- What controlled vocabulary options do we have for "role"? Currently only using creator.
- Should given name consist of first and middle name? That seems to be the international standard.
- Are we planning on supporting identifiers other than ORCID?

app/models/invenio_rdm_record_converter.rb
```diff
@@ -56,17 +56,52 @@ def versions(gf)
...
+      creators.map do |creator|
+        # dh_user = @known_creators.include?(creator) || (@known_creators << User.find_by(formal_name: creator))
+        # Galter Health Sciences Library is a User on production DigitalHub, but it does NOT have a formal_name value
+        dh_user = User.find_by(formal_name: creator)
+
+        if dh_user.present?
+          creator_type = "personal"
+          dh_user_display_name = dh_user.display_name.split(' ') # split name into components to be reused
+          family_name = dh_user_display_name.pop # remove last value from display name
+          given_name = dh_user_display_name.join(' ') # the remaining strings becomes given name
```
Meowcenary, line 84: I did not see anywhere to modify my display name on DH and my display name had my middle name in it so I think this should be okay to use here. It's easier than using formal name since we can just use `pop` to clean out the family name.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -56,17 +56,52 @@ def versions(gf)
 
     def invenio_metadata(gf)
       {
-        "resource_type": resource_type(gf)
+        "resource_type": resource_type(gf.resource_type.shift),
```
Meowcenary, line 59: Pass in the value needed directly instead of passing the entire file each time.

carsonicator:
To answer the three questions you had above:

> What controlled vocabulary options do we have for "role"? Currently only using creator.

In InvenioRDM, [creators](https://inveniordm.docs.cern.ch/reference/metadata/#creators-1-n) and [contributors](https://inveniordm.docs.cern.ch/reference/metadata/#contributors-0-n) both have a 'role' field. The list of controlled vocabularies for the different roles can be found [here](https://docs.google.com/spreadsheets/d/12Zk2Q2_Z6oGcWYBrmZB2UbwVN15FFdGmPnX5CdErYX8/edit#gid=758791748).

> Should given name consist of first and middle name? That seems to be the international standard.

Yes, see example for ‘creator’ [here](https://inveniordm.docs.cern.ch/reference/metadata/#creators-1-n).

> Are we planning on supporting identifiers other than ORCID?

The list of identifiers supported by InvenioRDM is [here](https://inveniordm.docs.cern.ch/reference/metadata/#identifiers-0-n). By default, it supports all identifiers included in [idutils](https://github.com/inveniosoftware/idutils).

@fenekku and Sara may have something else to add as well.

carsonicator (approved):
Thanks, Eric! I tried to answer your questions in the conversation thread.

saragon02:
Thanks, Matt and Eric. I agree with all Matt's comments above. In addition both the Contributor Roles and list of Identifiers agreed upon for use at Galter are viewable in their final lists here: https://northwestern.box.com/s/r1ttihhczk0pvph8w3newq0yg8i8avkm. The list of Identifiers we chose to use locally is a subset of the overall project's list, and is available here: https://northwestern.box.com/s/k1mq4xxpa2oo5hyp9bq31ckwq6c7p4q9

app/models/invenio_rdm_record_converter.rb
```diff
@@ -56,17 +56,52 @@ def versions(gf)
...
+          affiliations = [] # only used for creator_type "personal"
+        end
+
+        {
+          "name": dh_user.formal_name, # will always be this regardless of type
+          "type": creator_type,
+          "role": "creator", # TODO: roles are from custom controlled vocab, for now use this
+          "given_name": given_name, # TODO: seemingly the international standard is first and middle, but America seems to just use first
+          "family_name": family_name,
+          "identifiers": identifiers, # TODO: other than ORCID what identifiers do we need to support? can we lookup users for identifiers on import?
```
sharpattack, line 102: DigitalHub just has orcid from the list of supported creator identifier schemes in invenioRDM.

sharpattack (commented):
I think this will have to wait until we get a list back from our metadata experts of values from the creators that are "organisational".

app/models/invenio_rdm_record_converter.rb
```diff
@@ -56,17 +56,52 @@ def versions(gf)
...
         "type": irdm_type,
         "subtype": irdm_subtype
       }
     end
+
+    def creators(creators)
+      creators.map do |creator|
+        # dh_user = @known_creators.include?(creator) || (@known_creators << User.find_by(formal_name: creator))
+        # Galter Health Sciences Library is a User on production DigitalHub, but it does NOT have a formal_name value
+        dh_user = User.find_by(formal_name: creator)
```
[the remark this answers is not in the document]
Meowcenary: Good call. I'll check that out.

app/models/invenio_rdm_record_converter.rb
Meowcenary: That's what it seemed like to me. There were not a ton of users with that ID listed though and it seems to be an optional field for a user's profile. Maybe on the invenio import side of things it will make sense to try and look up the desired identifiers based on professional email or something?

fenekku:
TLDR; The InvenioRDM default role equivalency document is here: https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/vocabularies/contributor_role.csv 

There might be confusion here, so I will try to dispel it, but this is an illustration that falls into our communication challenges meeting.

The role values linked above are the *human readable* values. When importing / interacting with the API, we have to deal with *machine readable* values. The machine values are not in the linked document referred in another comment. They have yet to be made up. An example of such human/machine equivalencies is defined for our resource type vocabulary: https://github.com/galterlibrary/digitalhub-2/blob/master/app_data/vocabularies/resource_types.csv . The InvenioRDM default role equivalency document is here: https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/vocabularies/contributor_role.csv . The values you want to use are from the first column. For other such vocabularies, you can look in the parent folder: https://github.com/inveniosoftware/invenio-rdm-records/tree/master/invenio_rdm_records/vocabularies .

The confusion is a general issue that I have found to come up again and again so I will try to clear things up now. :warning: this might get somewhat philosophical...

What people read and what most of our metadata working group discussions revolve around are human readable values. Although we talk a lot about "standards" and source those from official controlled vocabularies, there isn't a functional component to those. Plus, we know what happens with standards...



The (pardon the pun) value in human readable values is really in how communities recognize them. We pick certain vocabularies because the community is familiar with those. To be precise, what they are familiar with is what these human readable values represent. So the bottom line real gold mine -- because we will never have a universal standard for anything -- is in the equivalency across standards/representations.

But then, how does one talk about the thing that should be equivalent to something else in another context? If I want to talk about "English" in a different contexts (e.g. in German) I need to have a hold on something that is not the contextualized / human readable word.   

 
`[Editor](human readable)   ------refers to / represents--------> [the concept of an editor](needs a machine representation)`

We need to give an identifier to `[the concept of an editor]`. This is what the machine readable value is.

Now that we have butchered a lot of philosophy to explain things, we can go back to the original problem and ask: "well, why not use 'Conceptualization role' as the machine identifier?" Mostly because it is a "bad" encoding: is it "conceptualization role", or "Conceptualization Role", why is it so english... The space makes it seem like a human readable value so causes confusion across the board. Not to mention that machine values can take advantage of not being seen by humans to encode information in their format.

Bottom line: for migration, whenever https://inveniordm.docs.cern.ch/reference/metadata mentions controlled vocabularies, we are talking about machine readable values and those can be found in the link above (and soon in https://github.com/inveniosoftware/invenio-vocabularies ).

Meowcenary:
Thank you for all the feedback everyone. I'm going to go through and clean things up now and hopefully get an update in later today.

Meowcenary:
I will have some questions for tomorrow, but I think for today I can focus on some other stuff to wrap up. I decided to just go through the content creators on DH to figure out what the known and unknown organizations were.

Meowcenary:
Oh no, I just though of another case I hadn't fully considered >_<

Meowcenary:
There's a significant amount of logical branches in this function so I think it will be worth it at some point to add a unit test even though it's private. At this point I'm taking a break from this particular subfield since I'm starting to lost it a little. I'm going to take on an ostensibly easy field (title) and then loop around to unit testing. IMO this is good enough to merge for now given that we have approaching deadlines and many fields to get through still.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -46,6 +60,74 @@
...
+    it do
+      is_expected.to eq json
+    end
+  end
+
+  let(:converter) { InvenioRdmRecordConverter.new }
+  let(:non_user_creator_name) { "I Don't Exist" }
+  let(:personal_creator_without_user_json) {
+    {
+      "creators": [{
```
Meowcenary, line 72: We could probably try to refactor this a bit to save lines, but for a small test it seemed best to keep it simple and clear by outlining exactly what the structure was for each. Also, this is probably going to have to be updated at some point.

Meowcenary:
@sharpattack added some more tests. Go ahead and give it a look when you get a chance. Second PR incoming!

app/models/invenio_rdm_record_converter.rb
```diff
@@ -5,11 +5,14 @@
...
 class InvenioRdmRecordConverter < Sufia::Export::Converter
   include Galtersufia::GenericFile::InvenioResourceTypeMappings
+  include Galtersufia::GenericFile::KnownOrganizations
 
   # Create an instance of a InvenioRdmRecordConverter converter containing all the metadata for json export
   #
   # @param [GenericFile] generic_file file to be converted for export
-  def initialize(generic_file)
+  def initialize(generic_file=nil)
+    return unless generic_file
```
Meowcenary, line 14: Defaulting the argument to nil was done so that a InvenioRdmRecordConverter instance could be created within the spec without having to supply it a generic file which would call all these functions. With this the spec can just test the creators function. 

I considered moving the creators function somewhere else, but it just doesn't seem to have anywhere else it belongs and especially since we've established a pattern of calling private methods to build out the export json.

spec/models/invenio_rdm_record_converter_spec.rb
sharpattack: No big deal for now. I do agree we should start organizing this spec as it grows. It would be nice to refactor this and add more context/describe blocks.

sharpattack (approved):
Looks good to me. Thank you for doing the extra "homework" needed to complete this. I appreciate the added tests, too.

Meowcenary:
Thanks for the follow up review @sharpattack! Yeah, we can always make it nice after we get the requirements out of the way.

PR #814, opened 2021-01-15
792 export metadata subfields title and additional titles

Meowcenary:
This branch has been rebased off of here: #812 so it would be best to hold off on reviewing this until that is merged. Once it is merged this PR will show only the title fields. I just want to make sure I get everything up and ready to go now so I don't get distracted with something else and forget before the long weekend.

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -77,7 +77,9 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   def invenio_metadata(gf)
     {
       "resource_type": resource_type(gf.resource_type.shift),
-      "creators": creators(gf.creator)
+      "creators": creators(gf.creator),
+      "title": gf.title.first,
+      "additional_titles": gf.title.last(gf.title.size-1).map{ |title| {"title": title, "type": "alternative_title", "lang": "eng"} }
     }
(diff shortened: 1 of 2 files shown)


app/models/invenio_rdm_record_converter.rb
```diff
@@ -77,7 +77,9 @@ def invenio_provenance(proxy_depositor, on_behalf_of)
   def invenio_metadata(gf)
     {
       "resource_type": resource_type(gf.resource_type.shift),
-      "creators": creators(gf.creator)
+      "creators": creators(gf.creator),
+      "title": gf.title.first,
+      "additional_titles": gf.title.last(gf.title.size-1).map{ |title| {"title": title, "type": "alternative_title", "lang": "eng"} }
```
sharpattack, line 82: Looks pretty straightforward to me.

sharpattack (approved):
Looks good, thank you for adding `title` and `additional_titles`

carsonicator (approved):
Thanks, @Meowcenary

app/models/invenio_rdm_record_converter.rb
Meowcenary: yeah, it was. I think I forgot to rebase this previously, but now it has been. Merging now.

PR #816, opened 2021-01-21
[#795] Export Metadata Subfield Subjects

### LINKED ISSUE #795: Export Metadata Subfield Subjects

Meowcenary:
closes #795

### DIFF
--- a/app/controllers/custom_authorities_controller.rb
+++ b/app/controllers/custom_authorities_controller.rb
@@ -15,7 +15,7 @@ class CustomAuthoritiesController < ApplicationController
     render json: hits
   end
 
-  def lcsh_names
+  def lcnaf_names
     lcnaf_base_uri = "http://id.loc.gov/authorities/names/suggest/?q="
     if params[:q].present?
       hits = JSON.parse(HTTParty.get(
@@ -23,7 +23,7 @@ class CustomAuthoritiesController < ApplicationController
       ))
       hits = hits.try(:[], 1) ? hits[1] : ["Error Searching"]
     end
-    
+
     render json: hits
   end
 
--- a/app/models/concerns/galtersufia/generic_file/header_lookup.rb
+++ b/app/models/concerns/galtersufia/generic_file/header_lookup.rb
@@ -1,0 +1,95 @@
+module Galtersufia
+  module GenericFile
+    module HeaderLookup
+      extend ActiveSupport::Concern
+
+      BASE_SPARQL_MESH_URI ="https://id.nlm.nih.gov/mesh/sparql?format=JSON&limit=10&inference=true&query=PREFIX%20rdfs"\
+        "%3A%20%3Chttp%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23%3E%0D%0APREFIX%20meshv%3A%20%3Chttp%3A%2F%2Fid.nlm."\
+        "nih.gov%2Fmesh%2Fvocab%23%3E%0D%0APREFIX%20mesh2018%3A%20%3Chttp%3A%2F%2Fid.nlm.nih.gov%2Fmesh%3E%0D%0A%0D%0ASELECT"\
+        "%20%3Fd%20%3FdName%0D%0AFROM%20%3Chttp%3A%2F%2Fid.nlm.nih.gov%2Fmesh%3E%0D%0AWHERE%20%7B%0D%0A%20%20%3Fd%20a%20"\
+        "meshv%3ADescriptor%20.%0D%0A%20%20%3Fd%20rdfs%3Alabel%20%3FdName%0D%0A%20%20FILTER(REGEX(%3FdName%2C%27"
+      END_SPARQL_MESH_URI = "%27%2C%20%27i%27))%20%0D%0A%7D%20%0D%0AORDER%20BY%20%3Fd%20%0D%0A"
+      LCSH_BASE_URI = "http://id.loc.gov/authorities/subjects/suggest/?q="
+
+      def pid_lookup_by_scheme(term, scheme)
+        scheme == :mesh ? mesh_term_pid_lookup(term) : lcsh_term_pid_lookup(term)
+      end
+
+      # return PID for provided mesh_header using SPARQL query
+      def mesh_term_pid_lookup(mesh_term="")
+        return unless mesh_term.present?
+
+        result = memoized_mesh_lookups[mesh_term]
+        return result if result.present?
+
+        query_result  = HTTParty.get(BASE_SPARQL_MESH_URI + mesh_term + END_SPARQL_MESH_URI)
+        json_parsed_result = JSON.parse(query_result)
+
+        if hits = hits_from_mesh_sparql(json_parsed_result)
+          result = pid_from_mesh_hits(hits)
+          memoized_mesh_lookups[mesh_term] = result
+          result
+        else
+          ["Error Searching"]
+        end
+      end
+
+      # return PID for provided lcsh_term
+      def lcsh_term_pid_lookup(lcsh_term="")
+        return unless lcsh_term.present?
+
+        result = memoized_lcsh_lookups[lcsh_term]
+        return result if result.present?
+
+        subject_names, subject_id_uris = perform_and_parse_lcsh_query(lcsh_term)
+
+        if subject_names && subject_id_uris
+          result = pid_from_lcsh_hits(lcsh_term, subject_names, subject_id_uris)
+          memoized_lcsh_lookups[lcsh_term] = result
+        else
+          ["Error Searching"]
+        end
+      end
+
+      private
+
+      def perform_and_parse_lcsh_query(lcsh_term)
+        # perform search with lcsh_term's whitespace replaced with '*' character
+        query_result = HTTParty.get(LCSH_BASE_URI + "*#{lcsh_term.gsub(/\s/,'*')}*")
+        hits = JSON.parse(query_result)
+
+        subject_names = hits.try(:[], 1)
+        subject_id_uris = hits.try(:[], -1)
+
+        return subject_names, subject_id_uris
+      end
+
+      def pid_from_lcsh_hits(lcsh_term, subject_names, subject_id_uris)
+        subject_match_index = subject_names.index{ |name| name == lcsh_term }
+        subject_id_uris[subject_match_index].split('/').last
+      end
+
+      def hits_from_mesh_sparql(json_parsed_result)
+        json_parsed_result.try(:[], "results").try(:[], "bindings")
+      end
+
+      def pid_from_mesh_hits(id_uris)
+        if id_uris.length > 1
+          id_uris.map{ |id_uri| id_uri["d"]["value"].split('/').last }
+        else
+          id_uris.shift["d"]["value"].split('/').last
+        end
+      end
+
+      # store searched mesh term as key and PID as value
+      def memoized_mesh_lookups
+        @memoized_mesh ||= {}
+      end
+
+      # store searched lcsh term as key and PID as value
+      def memoized_lcsh_lookups
+        @memoized_lcsh ||= {}
+      end
+    end
+  end
+end
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -6,6 +6,9 @@ include Sufia::Export
 class InvenioRdmRecordConverter < Sufia::Export::Converter
   include Galtersufia::GenericFile::InvenioResourceTypeMappings
   include Galtersufia::GenericFile::KnownOrganizations
+  include Galtersufia::GenericFile::HeaderLookup
+
+  SUBJECT_SCHEMES = [:tag, :mesh, :lcsh]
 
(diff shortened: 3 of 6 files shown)


app/models/invenio_rdm_record_converter.rb
```diff
@@ -138,4 +140,13 @@ def creators(creators)
       }
     end
   end
+
+  # return array of invenio formatted subjects
+  def subjects_for_scheme(terms, scheme)
+    if scheme != :tag
```
Meowcenary, line 146: I'm not a huge fan of this branch, but the only other options I could think of that would stay concise, which is what I liked about this solution, was to either merge in another hash (expensive, awkward) or delete the scheme key/value pair on each loop (awkward, maybe not expensive?).

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -42,17 +45,32 @@
...
-          "type": "alternative_title",
-          "lang": "eng"
-        },
-        {
-          "title": "Tertiary Title",
-          "type": "alternative_title",
-          "lang": "eng"
-        }],
-        "formats": "application/pdf"
+        "additional_titles": [
```
Meowcenary, line 48: I changed this formatting a little, but I can change it back. I keep flip flopping on this... JSON always looks kind of nasty to me for whatever reason.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -6,6 +6,7 @@
 class InvenioRdmRecordConverter < Sufia::Export::Converter
   include Galtersufia::GenericFile::InvenioResourceTypeMappings
   include Galtersufia::GenericFile::KnownOrganizations
+  SUBJECT_SCHEMES = [:tag, :mesh, :lcsh]
```
Meowcenary, line 9: This might be going over the top, but I figure I would normally use a constant in this sort of situation so why not use it here.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -42,17 +45,32 @@
...
+            "type": "alternative_title",
+            "lang": "eng"
+          },
+          {
+            "title": "Tertiary Title",
+            "type": "alternative_title",
+            "lang": "eng"
+          }
+        ],
+        "subjects": [
```
Meowcenary, line 60: As you will notice in here, the `identifier` field is left out entirely. I could not find a mapping quickly so I decided to get down what I could and then check back to see if there was somewhere to look these values up. [The documentation](https://inveniordm.docs.cern.ch/reference/metadata/#subjects-0-n) for this field listed it as optional (cardinality 0-1). @fenekku do you think this is something we should add to the export or can it be handled during import? I've got to believe there is some API for looking of LCSH and MESH terms, but I have no idea what to do for the keywords which are taken from GenericFile's tag field.

fenekku: I think this is something that should be done by the export if possible. DigitalHub has a local copy of MeSH terms (iirc) so it can use it to get the right identifier (it may also have local LCSH). For non-controlled terms a.k.a "keywords" those also go in subjects in Invenio but their scheme can be left empty.

All that being said, Lars is going over the vocabularies backend in InvenioRDM this month. Vocabularies are our generalization of the mechanism used for subjects terms. Do the best you can with the subjects for now, then mark it as blocked or make another ticket to revisit. We will have to revisit, once Lars' work is done to make sure we have the right structure. Let me know if that makes sense. We can take it to chat.

Meowcenary:
I was able to find a [MeSH lookup API](https://id.nlm.nih.gov/mesh/swagger/ui#/) that seems to fit what we'll need, but I'm still looking around for a LCSH lookup API. This is going to add significantly to the run time, but I'm not really sure what other options we have other than downloading the terms that exist and adding that to the project which would mean a significant amount of storage and I'm not really sure how much faster that would be.

app/models/concerns/galtersufia/generic_file/header_lookup.rb
```diff
@@ -0,0 +1,76 @@
...
+        end
+      end
+
+      private
+
+      def pid_from_mesh_hits(hits)
+        # TODO identify situations where there will be multiple "hits" - pids that match the search
+        # result = hits.map{|x| x["d"]["value"].split('/').last}
+        # below is solution for singlular "hit"
+        hits.shift["d"]["value"].split('/').last
```
Meowcenary, line 58: I really hate this, but I'm not really sure how else you can go about accessing such deeply nested values. At least it's wrapped in a function I guess.

app/models/concerns/galtersufia/generic_file/header_lookup.rb
```diff
@@ -0,0 +1,76 @@
...
+        if hits = hits.try(:[], "results").try(:[], "bindings")
+          result = pid_from_mesh_hits(hits)
+          memoized_mesh_lookups[mesh_term] = result
+          result
+        else
+          ["Error Searching"]
+        end
+      end
+
+      def lcsh_term_lookup(lcsh_term="")
```
Meowcenary, line 34: This is going to need more work as it's not returning the PID just yet. I'm trying to structure this similarly to `#mesh_term_lookup` so that it's both familiar in the process it takes to return a PID and if we feel like refactoring to something more abstract we can.

app/models/concerns/galtersufia/generic_file/header_lookup.rb
```diff
@@ -0,0 +1,76 @@
...
+      def mesh_term_lookup(mesh_term="")
+        return unless mesh_term.present?
+
+        result = memoized_mesh_lookups[mesh_term]
+        return result if result.present?
+
+        query_result  = HTTParty.get(BASE_SPARQL_MESH_URI + mesh_term + END_SPARQL_MESH_URI)
+        hits = JSON.parse(query_result)
+
+        if hits = hits.try(:[], "results").try(:[], "bindings")
```
Meowcenary, line 25: This should probably be wrapped into a function so that it has a semantic name.

app/models/concerns/galtersufia/generic_file/header_lookup.rb
```diff
@@ -0,0 +1,76 @@
+module Galtersufia
+  module GenericFile
+    module HeaderLookup
+      extend ActiveSupport::Concern
+
+      BASE_SPARQL_MESH_URI ="https://id.nlm.nih.gov/mesh/sparql?format=JSON&limit=10&inference=true&query=PREFIX%20rdfs"\
```
Meowcenary, line 6: Believe it or not this is actually a legitimate way to have a multiline string in Ruby. Go figure.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -46,6 +49,11 @@ def initialize(generic_file=nil)
     # @permissions = permissions(generic_file)
   end
 
+  def to_json(options={})
```
Meowcenary, line 52: The memoized header lookups are in a member variable so when `#to_json` is called it would include those in the return value. This keeps those values out.

app/models/concerns/galtersufia/generic_file/header_lookup.rb
```diff
@@ -0,0 +1,95 @@
...
+        hits = JSON.parse(query_result)
+
+        subject_names = hits.try(:[], 1)
+        subject_id_uris = hits.try(:[], -1)
+
+        return subject_names, subject_id_uris
+      end
+
+      def pid_from_lcsh_hits(lcsh_term, subject_names, subject_id_uris)
+        subject_match_index = subject_names.index{ |name| name == lcsh_term }
```
Meowcenary, line 68: Because the LCSH search was returning multiple values, even with the name search, this will first check the names array to find the index of the matching value. Once that index is found, the ids array is then indexed into to get the URI string that contains the PID.

app/models/concerns/galtersufia/generic_file/header_lookup.rb
```diff
@@ -0,0 +1,95 @@
...
+        "meshv%3ADescriptor%20.%0D%0A%20%20%3Fd%20rdfs%3Alabel%20%3FdName%0D%0A%20%20FILTER(REGEX(%3FdName%2C%27"
+      END_SPARQL_MESH_URI = "%27%2C%20%27i%27))%20%0D%0A%7D%20%0D%0AORDER%20BY%20%3Fd%20%0D%0A"
+      LCSH_BASE_URI = "http://id.loc.gov/authorities/subjects/suggest/?q="
+
+      def pid_lookup_by_scheme(term, scheme)
+        scheme == :mesh ? mesh_term_pid_lookup(term) : lcsh_term_pid_lookup(term)
+      end
+
+      # return PID for provided mesh_header using SPARQL query
+      def mesh_term_pid_lookup(mesh_term="")
```
Meowcenary, line 19: We might need to modify this for queries that return multiple values, but for now I think we have enough to work with.

app/models/concerns/galtersufia/generic_file/header_lookup.rb
```diff
@@ -0,0 +1,95 @@
...
+        return unless lcsh_term.present?
+
+        result = memoized_lcsh_lookups[lcsh_term]
+        return result if result.present?
+
+        subject_names, subject_id_uris = perform_and_parse_lcsh_query(lcsh_term)
+
+        if subject_names && subject_id_uris
+          result = pid_from_lcsh_hits(lcsh_term, subject_names, subject_id_uris)
+          memoized_lcsh_lookups[lcsh_term] = result
```
sharpattack, line 48: Do we need to return `result` here?

app/models/concerns/galtersufia/generic_file/header_lookup.rb
sharpattack: Whew, a lot of work just to get the PID for subject headings.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -144,4 +164,82 @@
...
+  let(:mesh_term) { "Vocabulary, Controlled" }
+  let(:mesh_query_url) do
+    "https://id.nlm.nih.gov/mesh/sparql?format=JSON&limit=10&inference=true&query=PREFIX%20rdfs%3A%20%3Chttp%3A%2F%2Fwww"\
+    ".w3.org%2F2000%2F01%2Frdf-schema%23%3E%0D%0APREFIX%20meshv%3A%20%3Chttp%3A%2F%2Fid.nlm.nih.gov%2Fmesh%2Fvocab%23%3E"\
+    "%0D%0APREFIX%20mesh2018%3A%20%3Chttp%3A%2F%2Fid.nlm.nih.gov%2Fmesh%3E%0D%0A%0D%0ASELECT%20%3Fd%20%3FdName%0D%0AFROM"\
+    "%20%3Chttp%3A%2F%2Fid.nlm.nih.gov%2Fmesh%3E%0D%0AWHERE%20%7B%0D%0A%20%20%3Fd%20a%20meshv%3ADescriptor%20.%0D%0A%20%"\
+    "20%3Fd%20rdfs%3Alabel%20%3FdName%0D%0A%20%20FILTER(REGEX(%3FdName%2C%27Vocabulary, Controlled%27%2C%20%27i%27))%20%0"\
+    "D%0A%7D%20%0D%0AORDER%20BY%20%3Fd%20%0D%0A"
+  end
+  let(:mesh_api_response) do
```
sharpattack, line 178: Not requesting the change in this PR, but to make you aware of this just in case: you could probably use vcr to record a cassette here.

sharpattack (commented):
Great job using memoized mesh and lcsh. Everything looks good, I just have two questions first before approving.

app/controllers/custom_authorities_controller.rb
```diff
@@ -15,15 +15,15 @@ def query_mesh
     render json: hits
   end
 
-  def lcsh_names
+  def lcnaf_names
```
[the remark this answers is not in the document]
Meowcenary: Oh, jeez, I didn't actually run the whole thing. I ran a search to check where this was called, but it's good to run the full suite.

app/models/invenio_rdm_record_converter.rb
Meowcenary: No problem, it was really odd to me that this was the standard way of doing this. At least it seemed to be the standard.

spec/models/invenio_rdm_record_converter_spec.rb
Meowcenary: I've had some problems in the past with VCR so I decided to just go with this. It's not a great reason to avoid VCR though so I should probably use it for the future.

app/models/concerns/galtersufia/generic_file/header_lookup.rb
Meowcenary: When you assign a value to a variable the return from that operation will be the value that was assigned. I think for the other function I explicitly call `result` to make it clear that it will return, but I guess I didn't here. I can change it though for consistency.

Meowcenary:
I forgot about this branch again. I think the best thing to do at this point is to close this PR, review locally, and then open a new PR that will be less cluttered.

PR #817, opened 2021-01-25
[#770] Export User Info W Deposits

### LINKED ISSUE #770: Export User Info
Output to csv file.

sharpattack:
Export script updated to include total deposits and last deposit date.
closes #770

### DIFF
--- a/lib/scripts/users_list.rb
+++ b/lib/scripts/users_list.rb
@@ -10,12 +10,15 @@ class UsersList
   def populate_user_info
     CSV.open(self.csv_file.path, "w") do |csv|
       # add headers
-      csv << ['Formal Name', 'NetID', 'Email', 'Current Sign In', 'Last Sign In']
+      csv << ['Formal Name', 'NetID', 'Email', 'Current Sign In', 'Last Sign In', 'Total Files Deposited', 'Last Deposit Date']
 
       select_all_users.find_in_batches do |users_batch|
         users_batch.each do |user|
+          solr_results = solr_query_by_depositor(user.username).sort_by { |result| result["date_uploaded_dtsi"] }
+          count = solr_results.count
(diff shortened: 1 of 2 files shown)


Meowcenary:
I'm going to wrap up the header lookup and then check this out. I think I have something "good enough" for now.

lib/scripts/users_list.rb
```diff
@@ -26,4 +29,16 @@ def select_all_users
     User.all.select(:id, :formal_name, :username, :email,
                     :current_sign_in_at, :last_sign_in_at)
   end
+
+  def solr_query_by_depositor(depositor)
```
Meowcenary, line 33: I like these tiny methods. They make it much easier to understand what's going on here.

lib/scripts/users_list.rb
```diff
@@ -26,4 +29,16 @@ def select_all_users
     User.all.select(:id, :formal_name, :username, :email,
```
Meowcenary, line 29: Why is this at the end? Is it for testing to verify the output?

Meowcenary (approved):
Looks good to me. Had some small questions, but nothing that should stop you from merging.

lib/scripts/users_list.rb
sharpattack: Do you mean the method itself? It was here for the same reason from your other comment: tiny methods to make it easier to understand. Or do you mean the `select` here? Or something else?

Meowcenary: I mean't why is it running `select` for all the users at the end? What is using the returned selection?

sharpattack: Ah, I see. `select` here just retrieves the certain fields I need, omitting all the other fields. https://apidock.com/rails/v4.2.7/ActiveRecord/QueryMethods/select

PR #820, opened 2021-01-29
[#795] Export Metadata Subfield Subjects

### LINKED ISSUE #795: Export Metadata Subfield Subjects

Meowcenary:
closes #795

### DIFF
--- a/app/controllers/custom_authorities_controller.rb
+++ b/app/controllers/custom_authorities_controller.rb
@@ -15,7 +15,7 @@ class CustomAuthoritiesController < ApplicationController
     render json: hits
   end
 
-  def lcsh_names
+  def lcnaf_names
     lcnaf_base_uri = "http://id.loc.gov/authorities/names/suggest/?q="
     if params[:q].present?
       hits = JSON.parse(HTTParty.get(
@@ -23,7 +23,7 @@ class CustomAuthoritiesController < ApplicationController
       ))
       hits = hits.try(:[], 1) ? hits[1] : ["Error Searching"]
     end
-    
+
     render json: hits
   end
 
--- a/app/models/concerns/galtersufia/generic_file/header_lookup.rb
+++ b/app/models/concerns/galtersufia/generic_file/header_lookup.rb
@@ -1,0 +1,92 @@
+module Galtersufia
+  module GenericFile
+    module HeaderLookup
+      extend ActiveSupport::Concern
+
+      BASE_SPARQL_MESH_URI ="https://id.nlm.nih.gov/mesh/sparql?format=JSON&limit=10&inference=true&query=PREFIX%20rdfs"\
+        "%3A%20%3Chttp%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23%3E%0D%0APREFIX%20meshv%3A%20%3Chttp%3A%2F%2Fid.nlm."\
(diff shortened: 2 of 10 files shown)


app/controllers/custom_authorities_controller.rb
```diff
@@ -15,15 +15,15 @@ def query_mesh
     render json: hits
   end
 
-  def lcsh_names
+  def lcnaf_names
```
Meowcenary, line 18: Changing this to reflect end point and what is being retrievd.

app/models/concerns/galtersufia/generic_file/header_lookup.rb
```diff
@@ -0,0 +1,92 @@
...
+        subject_id_uris[subject_match_index].split('/').last
+      end
+
+      def hits_from_mesh_sparql(json_parsed_result)
+        json_parsed_result.try(:[], "results").try(:[], "bindings")
+      end
+
+      def pid_from_mesh_hits(id_uris)
+        # take the first match found, split it's URI to get the PID at the end
+        id_uris.shift["d"]["value"].split('/').last
```
Meowcenary, line 78: I tested this a bit and the mesh hits matched better than the lcsh hits for whatever reason. It seems kind of pointless to sift through a list if we dont have to, but I guess we'll find out in testing if my understanding of this is incorrect.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -46,6 +49,11 @@ def initialize(generic_file=nil)
     # @permissions = permissions(generic_file)
   end
 
+  def to_json(options={})
```
Meowcenary, line 52: As I mentioned before, this is necessary to pass the options hash which in turn makes it so we can exclude certain variables from the object when we convert it to JSON format.

config/routes.rb
```diff
@@ -16,7 +16,7 @@
   get '/authorities/generic_files/creator' => 'custom_authorities#query_users'
   get '/authorities/generic_files/contributor' => 'custom_authorities#query_users'
   get '/authorities/generic_files/verify_user' => 'custom_authorities#verify_user'
-  get '/authorities/generic_files/subject_name' => 'custom_authorities#lcsh_names'
+  get '/authorities/generic_files/subject_name' => 'custom_authorities#lcnaf_names'
```
Meowcenary, line 19: Update to match the renamed controller method.

spec/controllers/custom_authorities_controller_spec.rb
```diff
@@ -195,21 +195,21 @@
     end
   end # query_users
 
-  describe '#lcsh_names', :vcr do
+  describe '#lcnaf_names', :vcr do
```
Meowcenary, line 198: Updating this spec for renamed method and route.

Meowcenary:
This is now updated and good to go. Hopefully this PR is much easier to read.

app/models/concerns/galtersufia/generic_file/header_lookup.rb
carsonicator: Just curious...so the ORM is translating requests into SPARQL queries?

spec/controllers/custom_authorities_controller_spec.rb
sharpattack: Great! Thanks for doing this.

sharpattack (approved):
Thanks for updating. This PR does not look much different than my previous review. If the full spec was green, this is good to me.

carsonicator (commented):
Thanks, @Meowcenary

PR #821, opened 2021-01-29
[#804] Export Metadata Subfield Locations

### LINKED ISSUE #804: Export Metadata Subfield Locations
Export the Metadata subfield [Locations](https://inveniordm.docs.cern.ch/reference/metadata/#locations-0-n)

Meowcenary:
closes #804

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -89,7 +89,8 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
       "title": gf.title.first,
       "additional_titles": gf.title.last(gf.title.size-1).map{ |title| {"title": title, "type": "alternative_title", "lang": "eng"} },
       "subjects": SUBJECT_SCHEMES.map{ |subject_type| subjects_for_scheme(gf.send(subject_type), subject_type) }.flatten,
-      "formats": gf.mime_type
+      "formats": gf.mime_type,
+      "locations": gf.based_near.present? ? gf.based_near.shift.split("', ").map{ |location| {place: location.gsub("'", "")} } : {}
     }
   end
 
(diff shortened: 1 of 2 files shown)


Meowcenary:
This was rebased off of #820 , so we should merge that first and then go through this.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -80,7 +88,9 @@ def invenio_metadata(gf)
       "creators": creators(gf.creator),
       "title": gf.title.first,
       "additional_titles": gf.title.last(gf.title.size-1).map{ |title| {"title": title, "type": "alternative_title", "lang": "eng"} },
-      "formats": gf.mime_type
+      "subjects": SUBJECT_SCHEMES.map{ |subject_type| subjects_for_scheme(gf.send(subject_type), subject_type) }.flatten,
+      "formats": gf.mime_type,
+      "locations": gf.based_near.present? ? gf.based_near.shift.split("', ").map{ |location| {place: location.gsub("'", "")} } : {}
```
Meowcenary, line 93: This was surprisingly annoying because of how the locations are entered. This gets the job done though even if it's ugly.

Meowcenary: FWIW I tested with both a single location and two locations, but I left it with two since splitting a string on a non-existent sub-string will return the full string in an array. In other words, `"some place, Illinois, USA".split("', ")` will return `["some place, Illinois, USA"]` which is then iterated over with map as expected.

sharpattack: Blech, that's how `based_near` stores it's values? Sorry, but thanks for dealing with it.

sharpattack (approved):
Simple enough, aside from all the shifting, splitting, mapping, and gsubbing. Thanks.

PR #823, opened 2021-02-02
[#793] Export Metadata Subfields Description And Additional Descriptions

### LINKED ISSUE #793: Export Metadata Subfields Description And Additional Descriptions
Export the Metadata subfields [Description](https://inveniordm.docs.cern.ch/reference/metadata/#description-1) and [Additional Descriptions](https://inveniordm.docs.cern.ch/reference/metadata/#additional-descriptions-0-n)

Meowcenary:
closes #793

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -88,6 +88,8 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
       "creators": creators(gf.creator),
       "title": gf.title.first,
       "additional_titles": gf.title.last(gf.title.size-1).map{ |title| {"title": title, "type": "alternative_title", "lang": "eng"} },
+      "description": gf.description.first,
+      "additional_descriptions": gf.description.last(gf.description.size-1).map{ |add_desc| {"description": add_desc, "type": "other", "lang": "eng"} },
       "subjects": SUBJECT_SCHEMES.map{ |subject_type| subjects_for_scheme(gf.send(subject_type), subject_type) }.flatten,
       "formats": gf.mime_type,
       "locations": gf.based_near.present? ? gf.based_near.shift.split("', ").map{ |location| {place: location.gsub("'", "")} } : {}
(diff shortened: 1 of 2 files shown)


app/models/invenio_rdm_record_converter.rb
```diff
@@ -88,6 +88,8 @@ def invenio_metadata(gf)
       "creators": creators(gf.creator),
       "title": gf.title.first,
       "additional_titles": gf.title.last(gf.title.size-1).map{ |title| {"title": title, "type": "alternative_title", "lang": "eng"} },
+      "description": gf.description.first,
```
Meowcenary, line 91: I'm a bit frustrated. The last time I looked at this I misunderstood how it was storing descriptions based on what I was reading back to CSV. I took it again today with file ids included to try and help understand the data better and realized it was storing descriptions as an array. I guess it work out in the end /shrug.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -88,6 +88,8 @@ def invenio_metadata(gf)
...
+      "additional_descriptions": gf.description.last(gf.description.size-1).map{ |add_desc| {"description": add_desc, "type": "other", "lang": "eng"} },
```
Meowcenary, line 92: I think leaving it as "other" to begin with will be fine. It's seemingly a required field so we need to put something, but really this will need to be looked over by a librarian I think.

Meowcenary:
Small and to the point.

app/models/invenio_rdm_record_converter.rb
sharpattack: This is an odd one, the way descriptions are stored. Anyway, what if there is only one description stored? Will this break?

Meowcenary: I'll check, but I think that `["entry"].last(0)` returns `[]` which is then `[].map{}` which again returns `[]`

Meowcenary: Sorry, forgot to reply. What I said above is the case.

sharpattack (approved):
Thanks for double checking what happens when there is one description. Small and to the point.

carsonicator (commented):
Thanks @Meowcenary

PR #824, opened 2021-02-03
[#795] Export Meta Subfield Subjects Redux

### LINKED ISSUE #795: Export Metadata Subfield Subjects

Meowcenary:
closes #795

### DIFF
--- a/app/models/concerns/galtersufia/generic_file/header_lookup.rb
+++ b/app/models/concerns/galtersufia/generic_file/header_lookup.rb
@@ -11,43 +11,39 @@ module Galtersufia
       END_SPARQL_MESH_URI = "%27%2C%20%27i%27))%20%0D%0A%7D%20%0D%0AORDER%20BY%20%3Fd%20%0D%0A"
       LCSH_BASE_URI = "http://id.loc.gov/authorities/subjects/suggest/?q="
 
-      def pid_lookup_by_scheme(term, scheme)
-        scheme == :mesh ? mesh_term_pid_lookup(term) : lcsh_term_pid_lookup(term)
+      def pid_lookup_by_scheme(term="", scheme="")
+        if term.blank? || scheme.blank?
+          return
+        elsif scheme == :mesh
+          memoized_mesh_lookups[term] || mesh_term_pid_lookup(term)
+        elsif scheme == :lcsh
+          memoized_lcsh_lookups[term] || lcsh_term_pid_lookup(term)
+        else
+          return
+        end
       end
 
       # return PID for provided mesh_header using SPARQL query
       def mesh_term_pid_lookup(mesh_term="")
-        return unless mesh_term.present?
+        hits = perform_and_parse_mesh_query(mesh_term)
 
-        result = memoized_mesh_lookups[mesh_term]
-        return result if result.present?
-
-        query_result  = HTTParty.get(BASE_SPARQL_MESH_URI + mesh_term + END_SPARQL_MESH_URI)
-        json_parsed_result = JSON.parse(query_result)
-
-        if hits = hits_from_mesh_sparql(json_parsed_result)
+        if hits.present?
           result = pid_from_mesh_hits(hits)
           memoized_mesh_lookups[mesh_term] = result
-          result
         else
-          ["Error Searching"]
+          nil
         end
       end
 
       # return PID for provided lcsh_term
       def lcsh_term_pid_lookup(lcsh_term="")
-        return unless lcsh_term.present?
-
-        result = memoized_lcsh_lookups[lcsh_term]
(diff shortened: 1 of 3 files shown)


app/models/concerns/galtersufia/generic_file/header_lookup.rb
```diff
@@ -11,43 +11,39 @@ module HeaderLookup
...
+        hits = perform_and_parse_mesh_query(mesh_term)
 
-        if hits = hits_from_mesh_sparql(json_parsed_result)
+        if hits.present?
           result = pid_from_mesh_hits(hits)
           memoized_mesh_lookups[mesh_term] = result
-          result
         else
-          ["Error Searching"]
+          nil
```
Meowcenary, line 34: Return nil if no results

app/models/concerns/galtersufia/generic_file/header_lookup.rb
```diff
@@ -11,43 +11,39 @@ module HeaderLookup
       END_SPARQL_MESH_URI = "%27%2C%20%27i%27))%20%0D%0A%7D%20%0D%0AORDER%20BY%20%3Fd%20%0D%0A"
       LCSH_BASE_URI = "http://id.loc.gov/authorities/subjects/suggest/?q="
 
-      def pid_lookup_by_scheme(term, scheme)
-        scheme == :mesh ? mesh_term_pid_lookup(term) : lcsh_term_pid_lookup(term)
+      def pid_lookup_by_scheme(term="", scheme="")
```
Meowcenary, line 14: refactor to save some lines

app/models/concerns/galtersufia/generic_file/header_lookup.rb
```diff
@@ -11,43 +11,39 @@ module HeaderLookup
...
+        if term.blank? || scheme.blank?
+          return
+        elsif scheme == :mesh
+          memoized_mesh_lookups[term] || mesh_term_pid_lookup(term)
```
Meowcenary, line 18: if the memoized result is found use it, otherwise return the function's return value.

app/models/concerns/galtersufia/generic_file/header_lookup.rb
```diff
@@ -11,43 +11,39 @@ module HeaderLookup
...
       def lcsh_term_pid_lookup(lcsh_term="")
-        return unless lcsh_term.present?
-
-        result = memoized_lcsh_lookups[lcsh_term]
-        return result if result.present?
-
         subject_names, subject_id_uris = perform_and_parse_lcsh_query(lcsh_term)
 
-        if subject_names && subject_id_uris
+        if subject_names.present? && subject_id_uris.present?
```
Meowcenary, line 42: I think this is the root of the original error. The query was actually returning something, but what we wanted from it was blank. This resolves the actual errors that ruby was throwing so we can do something else when there is no result.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -152,7 +152,15 @@ def creators(creators)
...
   def subjects_for_scheme(terms, scheme)
     if scheme != :tag
-      terms.map{ |term| {subject: term, identifier: pid_lookup_by_scheme(term, scheme), scheme: scheme} }
+      terms.map do |term|
+        pid = pid_lookup_by_scheme(term, scheme)
+
+        if pid.present?
+          {subject: term, identifier: pid, scheme: scheme}
+        else
+          {subject: "#{term}: DigitalHub field #{scheme}"}
```
Meowcenary, line 161: If there is no pid, mark it this way.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -222,26 +248,59 @@
...
+        @lcsh_pids = converter.lcsh_term_pid_lookup(lcsh_term)
+      end
+
+      it 'returns nil for PID for lcsh term' do
+        expect(@lcsh_pids).to eq(nil)
+      end
     end
+  end
+
+  let(:no_pid_mesh_subject) { [{subject: "Fake Lcsh Term: DigitalHub field mesh"}] }
```
Meowcenary, line 290: New tests for the specific behavior we want.

Meowcenary:
This is rebased off of the PR open for ticket 793 so that should be reviewed first.

sharpattack (approved):
Looks good, thank you for fixing. Whenever you get a chance, try deploying to staging and testing the export. If I recall, there was an LCSH lookup that broke the export on staging. This should handle that.

Meowcenary:
Thanks for checking it out. You're right, that was the reason for this new PR. BTW, thanks for finding that original bug hehe. Better to find it now than later on.

Issue #794, opened 2020-12-18
Export Metadata Subfields Publisher And Publication Date


Meowcenary:
Export the Metadata subfields [Publisher](https://inveniordm.docs.cern.ch/reference/metadata/#publisher-1) and [Publication Date](https://inveniordm.docs.cern.ch/reference/metadata/#publication-date-1)

PR #825, opened 2021-02-04
[#794] Export Metadata Subfields Publisher And Publication Date

Meowcenary:
Another simple and to the point PR. This has been rebase off of 795 Subfield Subjects Redux so review that first.

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -90,6 +90,8 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
       "additional_titles": gf.title.last(gf.title.size-1).map{ |title| {"title": title, "type": "alternative_title", "lang": "eng"} },
       "description": gf.description.first,
       "additional_descriptions": gf.description.last(gf.description.size-1).map{ |add_desc| {"description": add_desc, "type": "other", "lang": "eng"} },
+      "publisher": gf.publisher.shift,
+      "publication_date": "#{gf.date_uploaded.year}-#{gf.date_uploaded.month}-#{gf.date_uploaded.day}",
       "subjects": SUBJECT_SCHEMES.map{ |subject_type| subjects_for_scheme(gf.send(subject_type), subject_type) }.flatten,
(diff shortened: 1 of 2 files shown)


app/models/invenio_rdm_record_converter.rb
```diff
@@ -90,6 +90,8 @@ def invenio_metadata(gf)
       "additional_titles": gf.title.last(gf.title.size-1).map{ |title| {"title": title, "type": "alternative_title", "lang": "eng"} },
       "description": gf.description.first,
       "additional_descriptions": gf.description.last(gf.description.size-1).map{ |add_desc| {"description": add_desc, "type": "other", "lang": "eng"} },
+      "publisher": gf.publisher.shift,
```
Meowcenary, line 93: If no publisher present, `gf.publisher.shift` returns `nil` which responds to `to_json` with `"null"`. We can change this to something else if we want. Maybe an empty string?

Meowcenary:
Alrighty, I've cleared out some PRs so now I'm going to update this. It will be ready for review after that's done in a moment.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -90,6 +90,8 @@ def invenio_metadata(gf)
...
+      "publication_date": "#{gf.date_uploaded.month}-#{gf.date_uploaded.day}-#{gf.date_uploaded.year}",
```
sharpattack, line 94: For invenioRDM, should this be YEAR-MONTH-DAY?

sharpattack (approved):
Another simple PR, thanks. I will approve now for speed, but double-check the date format. I see it differently on the invenioRDM docs.

app/models/invenio_rdm_record_converter.rb
Meowcenary: You're right. I think I made this change before the follow up with just the `date_created` field. I'll fix this up before merging.

Meowcenary:
Running tests for this and if that looks good I'll merge, run the export on stage, and put the results on slack or in an email for everyone.

PR #826, opened 2021-02-08
[#797] Export Metadata Subfield Dates

### LINKED ISSUE #797: Export Metadata Subfield Dates
Export the Metadata subfield [Dates](https://inveniordm.docs.cern.ch/reference/metadata/#dates-0-n)

Meowcenary:
closes #797


Meowcenary:
As mentioned at standup and in the ticket, I ended up changing my approach to this from one that involved logic that could react to different date formats to one where I just made sure the original dates were formatted as we expected. There were not too many dates to be changed and I have a backup in place in case we need to restore any date formats that were changed. There are a few files without a value for `date_created` and one other case I wasn't sure about, but those can be changed on DigitalHub without a code change.

carsonicator (approved):
Thanks @Meowcenary

Meowcenary:
I'm going to cherry-pick this branch, re-push, and then merge. There's nothing to really review since it's just taking the `date_created` field without altering it.

Issue #828, opened 2021-02-18
Memoized Headers File


Meowcenary:
In order to save network calls and avoid errors related to repeatedly calling header lookup APIs, write out headers that are found to a file. When running the export read the values from this file.

PR #829, opened 2021-02-18
[#828] Memoized Headers File

Meowcenary:
Fix bugs with export, write out memoized headers to file, read memoized
headers from file before export. Update specs to reflect changes.


memoized_lcsh.txt
```diff
@@ -0,0 +1 @@
+{"Semantic Web"=>"sh2002000569"}
```
Meowcenary, line 1: I'm going to update this with what staging returned. I think it makes sense to just use this and remove the spec/fixtures files.

memoized_mesh.txt
```diff
@@ -0,0 +1 @@
+{"Vocabulary, Controlled"=>"D018875"}
```
Meowcenary, line 1: Same deal as LCSH file.

spec/fixtures/spec_memoized_lcsh.txt
```diff
@@ -0,0 +1 @@
+{"Semantic Web" => "sh2002000569"}
```
Meowcenary, line 1: As mentioned above, going to remove these.

spec/fixtures/spec_memoized_mesh.txt
```diff
@@ -0,0 +1 @@
+{"Vocabulary, Controlled" => "D018875"}
```
Meowcenary, line 1: Same as LCSH fixture.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -170,4 +170,16 @@ def subjects_for_scheme(terms, scheme)
       terms.map{ |term| {subject: term} }
     end
   end
+
+  def additional_titles(titles)
+    additional_titles_size = titles.size-1
+    return nil if additional_titles_size < 0
```
Meowcenary, line 176: It's annoying, but I did not want to use `shift` on the arrays for `description` and `title` here since that would change the actual structure of the array. By using this sort of messing select nothing is changed with the underlying array.

memoized_lcsh.txt
```diff
@@ -0,0 +1 @@
+{"Sailboats"=>"sh2011004125", "Periodical cicada"=>"sh85099887", "Saguenay River (Quebec)"=>"sh86006143", "Social stratification"=>"sh99000315", "Originality"=>"sh85095646", "Sailing"=>"sh91001077", "Marsden"=>"sh99014562", " Sam (Fictitious character)"=>"sh2008000533", "Molds (Fungi)"=>"sh85086566", "Lindens"=>"sh85077148", "Statues"=>"sh2003001401", "Test family"=>"sh85080045"}
```
Meowcenary, line 1: This isn't really a JSON file, it's just a ruby hash that we call `eval` on so I decided to just use the `txt` extension since it's versatile.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -15,6 +14,7 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   # @param [GenericFile] generic_file file to be converted for export
   def initialize(generic_file=nil)
     return unless generic_file
+    @@header_lookup = HeaderLookup.new
```
Meowcenary, line 17: Allocate the HeaderLookup as a class variable so it only needs to be allocated once. This should actually be changed to `||=` so I'll do that quick.

app/models/header_lookup.rb
```diff
@@ -0,0 +1,106 @@
...
+
+  def pid_from_mesh_hits(id_uris)
+    # take the first match found, split it's URI to get the PID at the end
+    id_uris.shift["d"]["value"].split('/').last
+  end
+
+  def strip_accents(term="")
+    # normalize with unicode normalization form kd, replace anything that
+    # is not a space, dash, parentheses, or ascii character with empty string
+    term.mb_chars.normalize(:kd).gsub(/[^\ \-()x00-\x7F]/n, '').to_s
```
Meowcenary, line 92: This is complicated, but basically "KD" was a standard for converting characters to and from ascii (that's what I was reading anyways). The regex is saying if it's not a space, dash, parentheses, or an ascii character with a hexadecimal value on the rang x00 to x7F then replace with an empty string.

app/models/header_lookup.rb
```diff
@@ -0,0 +1,106 @@
...
+  end
+
+  # read memoized from JSON file, if no file found return blank hash
+  def read_memoized_headers(filepath)
```
Meowcenary, line 96: Pretty simple read and eval function.

app/models/header_lookup.rb
```diff
@@ -0,0 +1,106 @@
...
+    "nih.gov%2Fmesh%2Fvocab%23%3E%0D%0APREFIX%20mesh2018%3A%20%3Chttp%3A%2F%2Fid.nlm.nih.gov%2Fmesh%3E%0D%0A%0D%0ASELECT"\
+    "%20%3Fd%20%3FdName%0D%0AFROM%20%3Chttp%3A%2F%2Fid.nlm.nih.gov%2Fmesh%3E%0D%0AWHERE%20%7B%0D%0A%20%20%3Fd%20a%20"\
+    "meshv%3ADescriptor%20.%0D%0A%20%20%3Fd%20rdfs%3Alabel%20%3FdName%0D%0A%20%20FILTER(REGEX(%3FdName%2C%27"
+  END_SPARQL_MESH_URI = "%27%2C%20%27i%27))%20%0D%0A%7D%20%0D%0AORDER%20BY%20%3Fd%20%0D%0A"
+  LCSH_BASE_URI = "http://id.loc.gov/authorities/subjects/suggest/?q="
+  MEMOIZED_MESH_FILE = "memoized_mesh.txt"
+  MEMOIZED_LCSH_FILE = "memoized_lcsh.txt"
+
+  def initialize(memoized_mesh_file_path=MEMOIZED_MESH_FILE, memoized_lcsh_file_path=MEMOIZED_LCSH_FILE)
+    @@memoized_mesh ||= read_memoized_headers(memoized_mesh_file_path)
```
Meowcenary, line 13: `||=` is a really helpful operator in ruby. It only assigns the value if it has not been assigned already.

sharpattack: I agree, I these are really great little operators.

memoized_lcsh.txt
```diff
@@ -0,0 +1 @@
+{"Sailboats"=>"sh2011004125", "Periodical cicada"=>"sh85099887", "Saguenay River (Quebec)"=>"sh86006143", "Social stratification"=>"sh99000315", "Originality"=>"sh85095646", "Sailing"=>"sh91001077", "Marsden"=>"sh99014562", " Sam (Fictitious character)"=>"sh2008000533", "Molds (Fungi)"=>"sh85086566", "Lindens"=>"sh85077148", "Statues"=>"sh2003001401", "Test family"=>"sh85080045", "Semantic Web"=>"sh2002000569"}
```
sharpattack, line 1: This and the `memoized_mesh.txt` results were from staging? Surprisingly small results to me. I expect production will be much larger. When we export on production, let's remember to get those results and commit it into the repository.

sharpattack (approved):
Thanks for fixing the header lookup. A lot of changes that was started by a small accent.

memoized_lcsh.txt
Meowcenary: Yeah, I was surprised too. I might run a little query to check myself on this, but I could see this being legit given it's basically a testing environment. If it seems like prod is isn't getting very many results we might need to dig deeper on this.

Meowcenary:
> Thanks for fixing the header lookup. A lot of changes that was started by a small accent.

The bug that launched a thousand lines of code lol.

PR #830, opened 2021-02-18
[#783] Export Files

### LINKED ISSUE #783: Export Invenio Files Field
**Feb 2021 Edit**

As of writing, all of the files metadata is actually generated by InvenioRDM when uploading a file. So this metadata doesn't have to be generated to export/upload the file. The actual JSON file needed is just this one from the REST API request: https://inveniordm.docs.cern.ch/reference/rest_api/#start-draft-file-uploads . That being said, given Katie's integrity requirement (really our own too) keeping track of the checksum so that we can compare it against the uploaded file would be needed. This could be kept track of in this extra JSON that ties things together.

sharpattack:
File metadata is not needed. Instead, we will use invenioRDM's Files
API to handle files metadata on upload. We just need the filename and
path of its contents. Changed the JSON output slightly to not include
the file information. closes #783


app/models/invenio_rdm_record_converter.rb
```diff
@@ -16,10 +16,34 @@ def initialize(generic_file=nil)
...
+
+  def filename_and_content_path(generic_file)
+    {
+      "filename": generic_file.filename,
+      "content_path": generic_file_content_path(generic_file.content.checksum.value)
+    }
+  end
+
+  def generic_file_content_path(checksum)
+    "/#{checksum[0..1]}/#{checksum[2..3]}/#{checksum[4..5]}/#{checksum}" unless !checksum
```
sharpattack, line 38: This is how contents' paths are generated; by taking the first 6 characters of it's checksum, and dividing it by 3. For example, a file with checksum of "abcd1234", is stored at the following path:
`ab/cd/12/abcd1234`

app/models/invenio_rdm_record_converter.rb
```diff
@@ -16,10 +16,34 @@ def initialize(generic_file=nil)
     return unless generic_file
     @@header_lookup ||= HeaderLookup.new
 
-    #PIDs
-    @pids = invennio_pids(generic_file.doi.shift)
-    @metadata = invenio_metadata(generic_file)
-    @provenance = invenio_provenance(generic_file.proxy_depositor, generic_file.on_behalf_of)
+    @record = record_for_export(generic_file)
+    @file = filename_and_content_path(generic_file)
+  end
```
sharpattack, line 21: This was so the JSON appears with the attributes/metadata and the file nested separately.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -16,10 +16,34 @@ def initialize(generic_file=nil)
...
+
+  def generic_file_content_path(checksum)
+    "/#{checksum[0..1]}/#{checksum[2..3]}/#{checksum[4..5]}/#{checksum}" unless !checksum
+  end
+
+  def record_for_export(generic_file)
+    {
+      "pids": invennio_pids(generic_file.doi.shift),
+      "metadata": invenio_metadata(generic_file),
+      "provenance": invenio_provenance(generic_file.proxy_depositor, generic_file.on_behalf_of)
```
sharpattack, line 45: We will continue adding the invenioRDM required fields here.

app/models/invenio_rdm_record_converter.rb
Meowcenary: I like it.

app/models/invenio_rdm_record_converter.rb
Meowcenary: Oh wow, that's nasty. It might be worthwhile to leave a comment here explaining that, but I can't think of any better way to deal with this.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -30,72 +30,78 @@
...
-      },
-      "provenance": {
-        "created_by": {
-          "user": assistant.username
-        },
-        "on_behalf_of": {
-          "user": user.username
         }
+      },
+      "file": {
```
Meowcenary, line 102: Good test. This makes it clear what the change does.

Meowcenary (approved):
Nice job figuring this out!

sharpattack:
@Meowcenary I added another two small commits, can you please re-review this PR whenever you have a moment?

Meowcenary (approved):
Thanks for the changes. Looks good to me!

PR #835, opened 2021-03-01
805 Export Metadata Subield Funding References

Meowcenary:
Add script to convert CSV of funding data into ruby hash stored within text file. Use function to read file and extract funding data.


spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -193,6 +207,24 @@
     end
   end
 
+  describe "#funding" do
```
Meowcenary, line 210: Some extra tests for the funding function.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -88,23 +90,35 @@
...
-        }
+        },
       },
       "file": {
         "filename": generic_file.filename,
         "content_path": nil # there is no file/content attached with this factory made GewnericFile
       }
     }.to_json
   end
+  let(:invenio_rdm_record_converter) { described_class.new(generic_file) }
```
Meowcenary, line 121: Putting this in a let variable so we can reuse a single instance for multiple specs.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -88,23 +90,35 @@
           ],
           "dates": [{"date": "1-1-2021", "type": "other", "description": "When the item was originally created."}],
           "formats": "application/pdf",
-          "locations": [{"place": "Boston, Massachusetts, United States"}, {"place": "East Peoria, Illinois, United States"}]
-        },
+          "locations": [{"place": "Boston, Massachusetts, United States"}, {"place": "East Peoria, Illinois, United States"}],
+          "funding": [{
```
Meowcenary, line 94: This is based on what was taken from the production csv and then converted into a hash stored in the text file above.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -11,6 +11,7 @@
   let(:generic_file) {
     make_generic_file(
       user,
+      id: "ns0646000",
```
Meowcenary, line 14: This was necessary for keying into the funding data.

prod_funding_sources.csv
```diff
@@ -0,0 +1,77 @@
+name,ror-identifier,award-number,file-id
```
Meowcenary, line 1: This is the data formatted for CSV.

funding_csv_to_hash.rb
```diff
@@ -0,0 +1,45 @@
+require 'csv'
```
Meowcenary, line 1: Simple script to write the CSV data into a hash which is then stored in a text fie.

funding_csv_to_hash.rb
```diff
@@ -0,0 +1,45 @@
...
+funding_csv.shift
+formatted_funding_data = funding_csv.each_with_object({}) do |row, hash|
+  # name, ror-identifier, award-number, file-id
+  name = row[0]
+  identifier = row[1]
+  award_number = row[2]
+  file_id = row[3]
+
+  if hash.keys.include?(file_id)
+    hash[file_id] +=
```
Meowcenary, line 13: I'm not sure if there is a better way to do this, but what's happening with this if/else is that if there is no key (file_id) in the hash, it will fail the first statement and then create an array with the funding data. If there is a key (file_id) in the hash, it will add to the array that's already there. I had an annoying bug with this were I was just overwriting the entries over and over when I needed to add to it for files with multiple funding sources.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -15,6 +15,7 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   def initialize(generic_file=nil)
     return unless generic_file
     @@header_lookup ||= HeaderLookup.new
+    @@funding_data ||= eval(File.read('app/models/concerns/galtersufia/generic_file/funding_data.txt'))
```
Meowcenary, line 18: Similar to what has been used for header lookup. Use a shared class variable and read from a hash stored in a text file with the magic of ruby's eval method.

app/models/concerns/galtersufia/generic_file/funding_data.txt
```diff
@@ -0,0 +1 @@
...r=>"K25HL119608", :identifier=>"", :scheme=>""}}, {:funder=>{:name=>"National Heart Lung and Blood Institute (NHLBI)", :identifier=>"012pb6c26", :scheme=>"ror"}, :award=>{:title=>"", :number=>"F30HL145995", :identifier=>"", :scheme=>""}}], "45a68a08-9aa6-4e13-8f7e-fd20f9d87010"=>[{:funder=>{:name=>"National Institute of Diabetes and Digestive and Kidney Diseases (NIDDK)", :identifier=>"00adh9b73", :scheme=>"ror"}, :award=>{:title=>"", :number=>"", :identifier=>"", :scheme=>""}}], "b93105d8-0142-4ddb-baa0-64fea8b981b2"=>[{:funder=>{:name=>"National Institute on Aging", :identifier=>"049v75w11", :scheme=>"ror"}, :award=>{:title=>"", :number=>" U2CAG057441", :identifier=>"", :scheme=>""}}, {:funder=>{:name=>"The Environmental Influences on Child Health Outcomes (ECHO)", :identifier=>"", :scheme=>"ror"}, :award=>{:title=>"", :number=>"", :identifier=>"", :scheme=>""}}, {:funder=>{:name=>"Office of the Director, National Institutes of Health (OD)", :identifier=>"00fj8a872", :scheme=>"ror"}, :award=>{:title=>"", :number=>"U24OD023319", :identifier=>"", :scheme=>""}}, {:funder=>{:name=>"", :identifier=>"", :scheme=>"ror"}, :award=>{:title=>"", :number=>"", :identifier=>"", :scheme=>""}}]}
```
Meowcenary, line 1: This is the file created from the script below

app/models/invenio_rdm_record_converter.rb
```diff
@@ -200,4 +202,8 @@ def additional_descriptions(descriptions)
     return nil if additional_descriptions_size < 0
     descriptions.last(additional_descriptions_size).map{ |add_desc| {"description": add_desc, "type": "other", "lang": "eng"} }
   end
+
+  def funding(file_id)
+    @@funding_data[file_id] || {}
```
Meowcenary, line 207: If I remember right we're no longer returning nil so I figured a blank hash would do.

prod_funding_sources.csv
sharpattack: Just curious, where did the source for these come from? I see at least one line with no data, and a few with just file-id. Why were they included here?

sharpattack (approved):
Looks good to me. Thanks for taking on this task, it doesn't look like it was an easy one.

prod_funding_sources.csv
Meowcenary: I took a excel file that Matt sent me and pulled out data that was needed for the metadata field. I'll attach that here so you can take a look. 

Honestly, we probably don't need to include the blank lines, but I figured it didn't hurt either and kept the function to access things consistent. No need to add a branch or anything.

Meowcenary:
Yeah, it was a little frustrating at times, but @carsonicator really helped me out with the excel file.

Meowcenary:
[DH_InvenioRDM_Funding_Data.xlsb.zip](https://github.com/galterlibrary/digital-repository/files/6086161/DH_InvenioRDM_Funding_Data.xlsb.zip) , sorry, took a bit to find it in my emails after I couldn't find it on my local.

PR #836, opened 2021-03-04
[#791] Export Metadata Subfield Creators - Redux

### LINKED ISSUE #791: Export Metadata Subfield Creators

Meowcenary:
Updating metadata subfield creators to reflect changes introduced with
recent InvenioRDM release. closes #791


app/models/invenio_rdm_record_converter.rb
```diff
@@ -133,42 +135,50 @@ def creators(creators)
...
-        identifiers = {}
-        affiliations = [] # only used for creator_type "personal"
-      # user
+        {
+          "person_or_org":
+            Hash.new.tap do |hash|
+              hash["name"] =  creator if creator.present? && !creator.include?("not identified")
+              hash["type"] = "organisational"
+            end
+        }.with_indifferent_access
```
Meowcenary, line 146: For some reason the keys were coming back as a mix of symbols and strings which was causing annoying issues with the spec for this function. The to_json spec is passing with this too so I think it's worth keeping in.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -44,15 +44,14 @@
...
-            "role": "",
-            "given_name": "#{user.display_name.split.first}",
-            "family_name": "#{user.display_name.split.last}",
-            "identifiers": {
-              "orcid": "#{user.orcid.split('/').last}"
-            },
-            "affiliations": []
+            "person_or_org": {
+              "type": "personal",
+              "given_name": "#{user.formal_name.split(',').last}",
```
Meowcenary, line 49: Formal name is really the field we should have been using this whole time. Display name is typically the same, but I could see a situation where someone uses a nickname or something similar.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -133,42 +135,50 @@ def creators(creators)
...
-        name = dh_user.formal_name # will always be this regardless of type
-        creator_type = "personal"
-        dh_user_display_name = dh_user.display_name.split(' ') # split name into components to be reused
-        family_name = dh_user_display_name.pop # remove last value from display name
-        given_name = dh_user_display_name.join(' ') # the remaining strings becomes given name
-        identifiers = dh_user.orcid.present? ? {"orcid": dh_user.orcid.split('/').pop} : {}
-        affiliations = [] # TODO: user.affiliation is nil for all users on production... where do we get this value?
-      # TODO: unknown personal, what do we enter for these fields?
+        dh_user_formal_name = dh_user.formal_name.split(',') # split name into components to be reused
+        family_name = dh_user_formal_name.shift # remove first value from formal name
```
Meowcenary, line 150: Mentioned this in the spec, but formal name is really the proper field to use here instead of display name.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -122,67 +121,58 @@
...
-          "identifiers": {},
-          "affiliations": []
-        }]
-    }
-  }.to_json
+        "person_or_org": {
+          "name": unidentified_creator_name
+        }
+      }]
+    }.with_indifferent_access
```
Meowcenary, line 145: There were some really annoying issues with keys coming back as a mix of symbols and strings. The to_json test is working as expected and all the values are what we expect so I think this is an okay solution to getting the spec working again.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -122,67 +121,58 @@
...
+  }
 
   describe "#creators" do
     context 'personal record without user in digital hub' do
       it 'assigns' do
-        expect({creators: converter.send(:creators, [non_user_creator_name])}).to eq(personal_creator_without_user_json)
+        expect({"creators": converter.send(:creators, [non_user_creator_name])}.with_indifferent_access).to eq(personal_creator_without_user_json)
```
Meowcenary, line 163: The indifferent_access is for the same reason above.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -133,42 +135,50 @@ def creators(creators)
...
+        given_name = dh_user_formal_name.join(' ') # the remaining strings becomes given name
+
+        {
+          "person_or_org":
+            Hash.new.tap do |hash|
+              hash["type"] = "personal"
+              hash["given_name"] = given_name
+              hash["family_name"] = family_name
+              hash["identifiers"] = {"orcid": dh_user.orcid.split('/').pop} if dh_user.orcid.present?
```
sharpattack, line 159: Shouldn't `identifiers` have the subfields `scheme` and `identifier`? Unless [this example](https://inveniordm.docs.cern.ch/reference/metadata/#__code_5) is out of date. It seems like `scheme` would only be orcid here. 

Also, what does this look like when there is no orcid for user? Empty string, nil, or just don't show this field?

app/models/invenio_rdm_record_converter.rb
```diff
@@ -133,42 +135,50 @@ def creators(creators)
...
-        identifiers = dh_user.orcid.present? ? {"orcid": dh_user.orcid.split('/').pop} : {}
-        affiliations = [] # TODO: user.affiliation is nil for all users on production... where do we get this value?
-      # TODO: unknown personal, what do we enter for these fields?
+        dh_user_formal_name = dh_user.formal_name.split(',') # split name into components to be reused
+        family_name = dh_user_formal_name.shift # remove first value from formal name
+        given_name = dh_user_formal_name.join(' ') # the remaining strings becomes given name
+
+        {
+          "person_or_org":
+            Hash.new.tap do |hash|
```
sharpattack, line 155: What about `name` subfield?

Meowcenary:
Going to rebase this and check for any breaking specs.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -133,42 +135,50 @@ def creators(creators)
...
-        given_name = ""
-        identifiers = {}
-        affiliations = [] # TODO: user.affiliation is nil for all users on production... where do we get this value?
+        family_name, given_name = creator.split(',')
+        {
+          "person_or_org":
+            Hash.new.tap do |hash|
+              hash["type"] = "personal"
+              hash["given_name"] = given_name.lstrip
+              hash["family_name"] = family_name.lstrip
```
sharpattack, line 178: What about `name` here too? Is that present if we have `given_name` and `family_name`?

app/models/invenio_rdm_record_converter.rb
sharpattack: Maybe this is in a release note somewhere, but is `name` subfield removed? if so, please ignore my other comments about adding it.

spec/models/invenio_rdm_record_converter_spec.rb
sharpattack: That's annoying. Could it be using `Hash` to create the result?(No need to refactor if it is. For now at least)

sharpattack (commented):
Overall, the redux looks great! Just want to confirm we're not missing `name` and that creator may not be "family_name, given_name".

app/models/invenio_rdm_record_converter.rb
```diff
@@ -133,42 +135,50 @@ def creators(creators)
...
-        creator_type = "organisational"
-        family_name = ""
-        given_name = ""
-        identifiers = {}
-        affiliations = [] # only used for creator_type "personal"
-      # user
+        {
+          "person_or_org":
+            Hash.new.tap do |hash|
+              hash["name"] =  creator if creator.present? && !creator.include?("not identified")
```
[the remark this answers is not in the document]
Meowcenary: Oh, good catch. I added a lot this stuff while debugging in a bit of a rush so I didn't even think of that.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -133,42 +135,50 @@ def creators(creators)
...
+        }
+      # Personal record without user in database
       else
-        name = creator
-        creator_type = "personal"
-        family_name = ""
-        given_name = ""
-        identifiers = {}
-        affiliations = [] # TODO: user.affiliation is nil for all users on production... where do we get this value?
+        family_name, given_name = creator.split(',')
```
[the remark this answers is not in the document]
Meowcenary: I think that formal name is always in the format of LastName, FirstName. I'll see if I can confirm this.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -133,42 +135,50 @@ def creators(creators)
...
+        {
+          "person_or_org":
+            Hash.new.tap do |hash|
+              hash["type"] = "personal"
```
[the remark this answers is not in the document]
Meowcenary: That was my line of reasoning too.

app/models/invenio_rdm_record_converter.rb
Meowcenary: Oh shoot, I overlooked that. 

I think for any field that does not have a value we just don't want to display it at all. I can confirm with @fenekku tomorrow at standup though.

app/models/invenio_rdm_record_converter.rb
Meowcenary: Screen Shot 2021-03-04 at 2 42 49 PM

Meowcenary: It's a little weird, but for a personal record it's not used. Organisational only.

app/models/invenio_rdm_record_converter.rb
Meowcenary: Same deal as above.

app/models/invenio_rdm_record_converter.rb
Meowcenary: Oh, I missed this top comment while going through the review. Mentioned it in other comments, but if the type is personal you do not use the name field.

app/models/invenio_rdm_record_converter.rb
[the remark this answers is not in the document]
Meowcenary: I looked into this more and formal name is populated by `NuLdap#standardized_name`. The spec for NuLdap `spec/models/nuldap_spec.rb` has all calls to `#standarized_name` return with a comma. I think from this it's safe to leave it as is.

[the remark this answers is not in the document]
sharpattack: Awesome, thanks for looking into this!

sharpattack (approved):
Looks good to go, thanks for your work on this @Meowcenary

carsonicator (commented):
Thanks @Meowcenary

PR #838, opened 2021-03-08
[#831] Resource Types Revisit

### LINKED ISSUE #831: Export resource types [revisit]
In the export excerpt and in the export zip so far, the resource types are not always filled in:

```json
    "resource_type": {
        "type": null,
        "subtype": "Pictorial Work, Photograph"
    },
```

The resource types should match up with the resource type conversion. See documents linked here: https://github.com/galterlibrary/digital-repository/issues/787#issuecomment-747463195 . 

So the above example should actually look like:

```json
    "resource_type": {
        "type": "image",
        "subtype": "image-pictorial_work"
    },
```

sharpattack:
Map the DigitalHub resource type to the correct InvenioRDM type and
subtype. closes #831


app/models/invenio_rdm_record_converter.rb
```diff
@@ -126,13 +126,23 @@ def invenio_metadata(gf)
...
-    {
-      "type": irdm_type,
-      "subtype": irdm_subtype
-    }
+    if irdm_types && irdm_types[1]
+      {
+        "type": irdm_types[0],
+        "subtype": irdm_types[1]
+      }
+    elsif irdm_types # only Dataset has no subtype
```
Meowcenary, line 136: Thanks for the comment here. I glossed over it at first and was looking for when this branch would run lol.

app/models/concerns/galtersufia/generic_file/invenio_resource_type_mappings.rb
```diff
@@ -4,177 +4,76 @@ module InvenioResourceTypeMappings
...
-        "Preprint": "Text Resources",
-        "Program": "Text Resources",
-        "Report": "Text Resources",
-        "Resource Guide": "Text Resources",
-        "Technical Documentation": "Text Resources",
-        # "Working Paper": "Text Resources",
-        "Bachelor Thesis": "Theses and Dissertations",
-        "Doctoral Thesis": "Theses and Dissertations",
-        "Masters Thesis": "Theses and Dissertations"
+      DH_IRDM_RESOURCE_TYPES = {
```
Meowcenary, line 7: Thanks for clearing this up and apologies for screwing it up so much the first time. I like the way you laid out the mapping here. Once I saw how you were using this it made a lot more sense to me why you took this approach.

Meowcenary (approved):
Looks good to me. Thanks for fixing this!

app/models/concerns/galtersufia/generic_file/invenio_resource_type_mappings.rb
sharpattack: Nothing was screwed up to me. I thought you did the mapping the right way the first time. As for changing the mapping, I wanted to keep the original, but I went with what was clear to me immediately. Looking back it may have turned out better due to at least one record not having a subtype.

PR #840, opened 2021-03-11
[#827] Series IV Collection

### LINKED ISSUE #827: Series IV Photographs Collection Changes
The Collection, [Series IV Photographs,](https://digitalhub.northwestern.edu/collections/x346d4254) has sub collections of only one item. 
- [ ] Move the item from each subcollection up one level to Series IV Photographs.
- [ ] Delete subcollection.

sharpattack:
For each subcollection's GenericFile, move it 'back' one collection
level to Series IV. The subcollection to be deleted after confirming
move. closes #827

### DIFF
--- a/lib/batch/fix_series_iv_827.rb
+++ b/lib/batch/fix_series_iv_827.rb
@@ -1,0 +1,17 @@
+# 827 Series IV Photographs Collection Changes
+# Move all records in each Series IV subcollection to Series IV collection,
+# and delete each subocllection.
+
+series_iv_collection = Collection.find("x346d4254")
+
+series_iv_collection.members.each do |subcollection|
+  if subcollection.class == Collection
+    subcollection.members.each do |member|
+      if member.class == GenericFile
+        series_iv_collection.members += [member]
+      end
+    end
+  end
+end
+
+series_iv_collection.save!


lib/batch/fix_series_iv_827.rb
```diff
@@ -0,0 +1,17 @@
+# 827 Series IV Photographs Collection Changes
+# Move all records in each Series IV subcollection to Series IV collection,
+# and delete each subocllection.
+
+series_iv_collection = Collection.find("x346d4254")
+
+series_iv_collection.members.each do |subcollection|
+  if subcollection.class == Collection
+    subcollection.members.each do |member|
+      if member.class == GenericFile
```
Meowcenary, line 10: What happens if there are other sub-collections within the sub-collection? E.g Collection -> Sub-collection1 -> [GenericFile1, GenericFile2, Sub-collection2 -> [GenericFile3, GenericFile4]]

Meowcenary (commented):
I think this is probably good to go, but I had one question before approval.

lib/batch/fix_series_iv_827.rb
sharpattack: That's a good question. Normally I would use recursion here in case the collections do continue to have other collections. For this particular one though, Series IV Photographs, there are no additional collection beyond this level so I didn't worry about adding recursion.

Meowcenary: I figured that might be the case. Thanks for following up on this and glad you had already considered it.

Meowcenary (approved):
Thanks for following up and for taking care of this!

PR #841, opened 2021-03-11
[#796] Export Metadata Subfield Contributors

### LINKED ISSUE #796: Export Metadata Subfield Contributors
Export the Metadata subfield [Contributors](https://inveniordm.docs.cern.ch/reference/metadata/#contributors-0-n)

**Feb 2021 EDIT**
This field has the same structure as the creators field ~~, but the role is required,~~ .So it's mostly a matter of completing the Creators task https://github.com/galterlibrary/digital-repository/issues/791  (and making sure a role is given for contributors).

Edit: Role is **NOT** required so this is the same as the fields for creators. In addition to filling the contributors fields I think this is a good chance to look at memoizing creators similarly to how headers are memoized.

Meowcenary:
Refactor and rename creators function to build json for either creators
or contributors. Memoize json for creators/contributors for future use.
closes #796


app/models/invenio_rdm_record_converter.rb
```diff
@@ -12,14 +12,17 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
...
 
   # Create an instance of a InvenioRdmRecordConverter converter containing all the metadata for json export
   #
   # @param [GenericFile] generic_file file to be converted for export
   def initialize(generic_file=nil)
     return unless generic_file
+
     @@header_lookup ||= HeaderLookup.new
     @@funding_data ||= eval(File.read('app/models/concerns/galtersufia/generic_file/funding_data.txt'))
+    @@person_or_org_data ||= eval(File.read(MEMOIZED_PERSON_OR_ORG_DATA_FILE))
```
Meowcenary, line 25: Class variable (shared across all instances of the class) that follows the same method used for reading funding data.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -109,14 +112,15 @@ def invenio_provenance(proxy_depositor, on_behalf_of)
   def invenio_metadata(gf)
     {
       "resource_type": resource_type(gf.resource_type.shift),
-      "creators": creators(gf.creator),
+      "creators": gf.creator.map{ |creator| build_creator_contributor_json(creator) },
```
Meowcenary, line 115: Refactoring the original creators function into #build_creator_contributor_json(creator) allowed for a simple map to be called like this.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -145,56 +149,58 @@ def resource_type(digitalhub_subtype)
...
-          "person_or_org":
-            Hash.new.tap do |hash|
-              hash["type"] = "personal"
-              hash["given_name"] = given_name.lstrip
-              hash["family_name"] = family_name.lstrip
-            end
-        }
-      end
+  def build_creator_contributor_json(creator)
+    if creator_data = @@person_or_org_data[creator]
```
Meowcenary, line 153: This is the key line for memoization.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -145,56 +149,58 @@ def resource_type(digitalhub_subtype)
...
+      return creator_data
+    # Organization
+    elsif organization?(creator)
+      json = @@person_or_org_data[creator] = {
```
Meowcenary, line 157: `json = ...` is done so that the actual json is returned at the end and not whatever is returned from the file write.

memoized_person_or_org_data.txt
```diff
@@ -0,0 +1 @@
+{"Tester, Mock"=>{:person_or_org=>{"type"=>"personal", "given_name"=>" Mock", "family_name"=>"Tester", "identifiers"=>{:scheme=>"orcid", :identifier=>"1234-5678-9123-4567"}}}, "User, Contributor"=>{:person_or_org=>{"type"=>"personal", "given_name"=>" Contributor", "family_name"=>"User"}}, "Laster, Firston"=>{:person_or_org=>{"type"=>"personal", "given_name"=>"Firston", "family_name"=>"Laster"}}, "Unknown"=>{:person_or_org=>{"name"=>"Unknown"}}, "Creator not identified."=>{:person_or_org=>{"name"=>"Creator not identified."}}, "Galter Health Sciences Library"=>{:person_or_org=>{"name"=>"Galter Health Sciences Library", "type"=>"organisational"}}}
```
Meowcenary, line 1: This is from the test data, but I figure it's a fine place to start with this.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -145,56 +149,58 @@ def resource_type(digitalhub_subtype)
...
-        {
-          "person_or_org":
-            Hash.new.tap do |hash|
-              hash["type"] = "personal"
-              hash["given_name"] = given_name.lstrip
-              hash["family_name"] = family_name.lstrip
-            end
-        }
-      end
+  def build_creator_contributor_json(creator)
```
fenekku, line 152: :warning: A subtlety to be aware of is that `contributors` require a `role` whereas `creators` don't (`role` is optional for `creators`). So either we need to distinguish between the 2 or always provide one.

Meowcenary:
I'm looking into roles for contributors now. I'm getting the feeling that there is no role being stored for contributors on DigitalHub, so we might need to come up with some generic term for contributor to work here.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -145,58 +150,68 @@ def resource_type(digitalhub_subtype)
...
-              hash["type"] = "personal"
-              hash["given_name"] = given_name.lstrip
-              hash["family_name"] = family_name.lstrip
-            end
-        }
-      end
+  def contributors(contributors)
+    contributors.map do |contributor|
+      contributor_json = build_creator_contributor_json(contributor)
+      contributor_json[:person_or_org].merge!({"role": ROLE_OTHER})
```
Meowcenary, line 156: Here is where the role is included for contributors per @fenekku's feedback.

sharpattack: What happens if there are more than one contributors? Ruby will return each `contributor_json`, or just the last `contributor_json`?

app/models/invenio_rdm_record_converter.rb
```diff
@@ -12,14 +12,18 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   ENGLISH = "english"
   CREATOR_UNKNOWN = 'unknown'
   CREATOR_NOT_IDENTIFIED = "creator not identified."
+  MEMOIZED_PERSON_OR_ORG_DATA_FILE = 'memoized_person_or_org_data.txt'
+  ROLE_OTHER = 'Other'
```
sharpattack: Needs to be lowercase
```suggestion
  ROLE_OTHER = 'other'
```

sharpattack (commented):
Looks good to me, just one question about multiple contributors. When we go through the upcoming release process soon, let's be sure to try this out on staging.

app/models/invenio_rdm_record_converter.rb
sharpattack: This is outdated. Github must have saved the comment before the change was updated.

app/models/invenio_rdm_record_converter.rb
sharpattack: Just learned that`#map` returns the array. Thanks for explaining on the side @Meowcenary

sharpattack (approved):
Okay, now that I understand `#map` better, this looks good. Also, I like that you recycled `#build_creator_contributor_json`.

PR #842, opened 2021-03-12
[#800] Metadata Subfield Sizes

### LINKED ISSUE #800: Export Metadata Subfield Sizes
Export the Metadata subfield [Sizes](https://inveniordm.docs.cern.ch/reference/metadata/#sizes-0-n)

Meowcenary:
Add sizes subfield to metadata function. Only page counts are included
at this time. closes #800

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -119,6 +119,7 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
       "subjects": SUBJECT_SCHEMES.map{ |subject_type| subjects_for_scheme(gf.send(subject_type), subject_type) }.flatten,
       "dates": gf.date_created.map{ |date| {"date": date, "type": "other", "description": "When the item was originally created."} },
       "languages": gf.language.any?{ |lang| lang.downcase == ENGLISH} ? ["eng"] : "",
+      "sizes": Array.new.tap{ |size_json| size_json << "#{gf.page_count} pages" if !gf.page_count.blank? },
       "formats": gf.mime_type,
       "locations": gf.based_near.present? ? gf.based_near.shift.split("', ").map{ |location| {place: location.gsub("'", "")} } : {},
       "funding": funding(gf.id)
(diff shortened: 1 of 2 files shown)


spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -28,7 +28,8 @@
       date_created: ["1-1-2021"],
       mime_type: 'application/pdf',
       grants_and_funding: ["European Commission 00k4n6c32"],
-      language: ["English"]
+      language: ["English"],
+      page_count: [rand(1..1000).to_s]
```
Meowcenary, line 32: Randomize the page count

app/models/invenio_rdm_record_converter.rb
```diff
@@ -119,6 +119,7 @@ def invenio_metadata(gf)
       "subjects": SUBJECT_SCHEMES.map{ |subject_type| subjects_for_scheme(gf.send(subject_type), subject_type) }.flatten,
       "dates": gf.date_created.map{ |date| {"date": date, "type": "other", "description": "When the item was originally created."} },
       "languages": gf.language.any?{ |lang| lang.downcase == ENGLISH} ? ["eng"] : "",
+      "sizes": Array.new.tap{ |size_json| size_json << "#{gf.page_count} pages" if !gf.page_count.blank? },
```
Meowcenary, line 122: Tap is just the greatest for this kind of thing.

Meowcenary:
Very small and simple PR. File size is no longer included since Invenio will calculate file size itself.

sharpattack (commented):
Looks good to me. Just curious though, why only `pages`? Do you have an example record with this, I can't seem to find one. Any other "size" info we could use?

Meowcenary:
Here is the [documentation on Sizes](https://inveniordm.docs.cern.ch/reference/metadata/#sizes-0-n) if you would like to see an official example. File size is not being used because, per @fenekku's feedback, it will be calculated by the database. The sufia model only had page counts and byte size so it seemed that page count was the only option that could go in. Also FWIW, this is an optional field so I think _anything_ we put in is just details.

fenekku:
On top of my head, the only other "sizes" is duration when the resource is a movie/audio clip, but don't worry about it.

sharpattack (approved):
Found an example in DigitalHub. I was only seeing "pages" on Collections, and thought there may not be any on GenericFiles. Looks good to go.

PR #846, opened 2021-03-19
[#844] Rework Series IV Script

### LINKED ISSUE #844: Rework the Script for Series IV Photographs Changes
The script from #827 will not work. I found out the items in the subcollections are `Page` objects, which inherits from `GenericFile` class. I will need to do a little more testing, because the subcollections are considered to be "multi-paged", giving their respective items a `combined_file_id` that probably won't be automatically updated. This could potentially break the Series IV Collection. This is not as straightforward and simple as initially thought.

sharpattack:
Items in the subcollection were Pages, and not GenericFile, and had
combined_file relation to the subcollection. Check if item is Page, and
remove the combined_file relation. closes #844

### DIFF
--- a/lib/batch/fix_series_iv_827.rb
+++ b/lib/batch/fix_series_iv_827.rb
@@ -3,14 +3,19 @@
 # and delete each subocllection.
 
 series_iv_collection = Collection.find("x346d4254")
+# series_iv_collection = Collection.find("fj236226r") # staging example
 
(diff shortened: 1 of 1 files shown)


Meowcenary (approved):
Good job figuring this out!

sharpattack:
Ran the script on staging, and is now working. Some manual work post run still needed, such as deleting the subcollections, but this is enough.

PR #848, opened 2021-03-19
[#753] Create List Of Netid, Email, Fullname From Owners Of Records

### LINKED ISSUE #753: Create list of netid, email, fullname from owners of records

Meowcenary:
Add script for exporting a CSV of file owner data. closes #753

### DIFF
--- a/export_dh_users_with_groups.rb
+++ b/export_dh_users_with_groups.rb
@@ -1,0 +1,10 @@
+file_name = "dh_users_with_groups.csv"
+
+CSV.open(file_name, "w") do |csv|
+  headers = ["Owner Net Id", "Owner Email", "Owner Fullname", "Owner Groups"]
+  csv << headers
+
+  User.all.each do |user|
+    csv << [user.username, user.email, user.name, user.groups]
+  end
+end


export_file_owners_as_csv.rb
```diff
@@ -0,0 +1,12 @@
+file_name = "dh_file_owners.csv"
+
+CSV.open(file_name, "w") do |csv|
+  headers = ["File Id", "Owner Net Id", "Owner Email", "Owner Fullname"]
+  csv << headers
+
+  GenericFile.all.each do |gf|
+    proxy_depositor = gf.proxy_depositor
+    user = proxy_depositor.present? ? User.find_by(username: proxy_depositor) : User.find_by(username: gf.depositor)
```
sharpattack, line 9: @fenekku checking with you on this: if there is a proxy_depositor, then just get their info? Ignore the depositor?

sharpattack (commented):
Looks good to me, just to want to clarify that we don't want depositor in addition to the proxy_depositor, if there is a proxy_depositor.

export_file_owners_as_csv.rb
fenekku: tldr; getting a list of *all* DH users is the simplest for this specific task. Read below for the bigger picture.

After some thought, here is the goal:

When importing the records, we need to have the information about "permission and visibility" for that record. Usually (and this is where I defer to your knowledge of DigitalHub), that would ultimately map to something like: this category of users can perform these actions (for different categories of users and different actions). Here we may want to discuss in person what are the equivalences between permissions in DH and permissions in InvenioRDM.

The most basic version of visibility and ownership is currently supported in InvenioRDM. When a record is created the following information can be passed to it:

```json
{
  "access": {
    "files": "public" | "restricted",
    "record": "public" | "restricted",
    "embargo": {
        "until": "...",
        "active": false,
        "reason": "..."
    }
  }
},
```
Marking a `files` or `record` as "restricted" makes it only visible to its owner. And the owner is determined by the authenticated API user for now. We haven't yet implemented the mechanism to assign permission to other users, so the export structure for that can't be implemented yet. However even if the export structure did exist, the ids in DH wouldn't match up the ids in InvenioRDM, so we would need to create a section in the json where the permissions are expressed with netids and the importer can retrieve the right InvenioRDM user ids corresponding to the DH netids. 

In any case, from the perspective of the export, what is important is to communicate the "permission and visibility" so that the connections can be made from the import side. If we have a list of all DH users, then down the road, no matter who shows up as an owner of a record in DH, there will be a mapping that the importer can use to retrieve the right user in InvenioRDM. 

Finally, and more directly to your question @sharpattack : if the meaning of owneship in DH includes proxy_depositors AND who they are proxying for, then in a separate future task, both of these should be encoded as owners. InvenioRDM supports multiple owners.

sharpattack (approved):
Apologies if this is not ready for re-review, but I figure to approve now. I can always review any additional changes. This looks good to me. Glad we clarified what we wanted, and went with a much more simplified result.

Meowcenary:
Oh, it was ready. I thought it would ping you somehow, but I should have been more clear on that. I'll shoot you a message next time.

PR #850, opened 2021-03-23
[#849] Creator/Contributor given_name fix

### LINKED ISSUE #849: Fix No given_name for Creator or Contributor
Tested on staging, but could be true on production. If a record's creator/contributor does not have a properly formatted name, then it will break at the last `else` branch of `#build_creator_contributor_json` when trying to get this info.

sharpattack:
Not all creator/contributor will be in the suggested "family_name,
given_name" format. If we encounter different formats, mark it as type
"organisational". Add type to all creator/contributor. closes #849


Meowcenary (approved):
Looks good. As usual props on the specs!

app/models/invenio_rdm_record_converter.rb
```diff
@@ -167,7 +165,7 @@ def build_creator_contributor_json(creator)
       json = @@person_or_org_data[creator] = {
         "person_or_org":
           Hash.new.tap do |hash|
-            hash["name"] =  creator if creator.present? && (!creator.include?(CREATOR_NOT_IDENTIFIED) || !creator.include(CREATOR_UNKNOWN))
+            hash["name"] =  creator
```
Meowcenary, line 168: Good job getting rid of this mess. Sorry for leaving that to you to clean up.

PR #851, opened 2021-03-23
[#803] Export Metadata Subfield Rights

### LINKED ISSUE #803: Export Metadata Subfield Rights
Export the Metadata subfield [Rights](https://inveniordm.docs.cern.ch/reference/metadata/#rights-0-n). 

At the time of this ticket being created, according to the documentation "this field may change after work done on controlled vocabulary management is completed during December 2020". With that said leave this ticket for after other subfields are created.

Meowcenary:
closes #803

### DIFF
--- a/app/models/concerns/galtersufia/generic_file/license_data.txt
+++ b/app/models/concerns/galtersufia/generic_file/license_data.txt
@@ -1,0 +1,78 @@
+{
+  'All Rights Reserved': {
+    "name": 'All Rights Reserved',
+    "licenseId": 'ARR'
+  },
+  'http://creativecommons.org/publicdomain/mark/1.0/': {
+    "name": 'Creative Commons Public Domain Mark 1.0',
+    "licenseId": 'CC-PDM-1.0'
+  },
+  'http://creativecommons.org/licenses/by-nc-sa/3.0/us/': {
+    "name": 'Creative Commons Attribution Non Commercial Share Alike 3.0 United States',
+    "licenseId": "CC-BY-NC-SA-3.0-US"
+  },
+  'http://creativecommons.org/licenses/by-nc-nd/3.0/us/': {
+    "name": 'Creative Commons Attribution Non Commercial No Derivatives 3.0 United States',
+    "licenseId": "CC-BY-NC-ND-3.0-US"
+  },
+  'http://creativecommons.org/licenses/by-nd/3.0/us/': {
+    "name": 'Creative Commons Attribution No Derivatives 3.0 United States',
+    "licenseId": "CC-BY-ND-3.0-US"
+  },
+  'http://creativecommons.org/licenses/by-nc/3.0/us/': {
+    "name": 'Createive Commons Attribution Non Commercial 3.0 United States',
+    "licenseId": "CC-BY-NC-3.0-US",
+  },
+  'http://creativecommons.org/licenses/by-sa/3.0/us/': {
+    "name": "Creative Commons Attribution Share Alike 3.0 United States",
+    "licenseId": "CC-BY-SA-3.0-US",
+  },
+  'http://creativecommons.org/licenses/by/4.0/': {
+    "reference": "./CC-BY-4.0.json",
+    "isDeprecatedLicenseId": false,
+    "detailsUrl": "./CC-BY-4.0.html",
+    "referenceNumber": 109,
+    "name": "Creative Commons Attribution 4.0 International",
+    "licenseId": "CC-BY-4.0",
+    "seeAlso": [
+    	"https://creativecommons.org/licenses/by/4.0/legalcode"
+    ],
+    "isOsiApproved": false
+  },
+  'http://creativecommons.org/licenses/by/3.0/us/': {
+    "reference": "./CC-BY-3.0-US.json",
+    "isDeprecatedLicenseId": false,
+    "detailsUrl": "./CC-BY-3.0-US.html",
+    "referenceNumber": 70,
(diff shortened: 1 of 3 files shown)


app/models/concerns/galtersufia/generic_file/license_data.txt
```diff
@@ -0,0 +1,77 @@
+{
+  'All Rights Reserved': {
```
Meowcenary, line 2: I need to ask about this, but I didn't think it should hold up opening a PR.

From what I understand this is a standard way of saying "No one can do anything with this item" and because of that there isn't a website, id, or really even a license so I'm not sure what goes here.

Meowcenary: Side note: I just copied over JSON directly from [SPDX's](https://spdx.org/licenses/) list except for the licenses I had to enter myself. Because of this there are a lot of unnecessary fields included, but I didn't think they were really hurting anything so I just left them.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -12,19 +12,23 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   ENGLISH = "english"
   CREATOR_UNKNOWN = 'unknown'
   CREATOR_NOT_IDENTIFIED = "creator not identified."
-  MEMOIZED_PERSON_OR_ORG_DATA_FILE = 'memoized_person_or_org_data.txt'
   ROLE_OTHER = 'other'
+  DEFAULT_RIGHTS_SCHEME = "spdx"
```
Meowcenary, line 16: Some very small refactoring here. I added a constant for one of the file paths and also moved the class variables from the initializer function to the class level. Moving the class variables out of the initializer function makes it easier to do unit testing since we no longer have to create an instance of the class to have these variables populate.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -245,4 +250,17 @@ def additional_descriptions(descriptions)
   def funding(file_id)
     @@funding_data[file_id] || {}
   end
+
+  def rights(license_urls)
```
Meowcenary, line 254: I don't think any of the files will have multiple licenses since it seemed like the intended workflow was that you would select a single license. However, because you can both shift+click and ctrl+click on different options in the list it's possible for a determined person to assign multiple licenses. For that reason I'm using a map loop here rather than just shifting the array to get the license url.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -279,6 +282,45 @@
...
+  let(:expected_mit) do
+    {
+      "rights": "MIT License",
+      "scheme": "spdx",
+      "identifier": "MIT",
+      "url": mit_license_url
+    }
+  end
+
+  describe "#rights" do
```
Meowcenary, line 316: It's not the most helpful spec, but I didn't think it hurt to add a few more examples.

app/models/concerns/galtersufia/generic_file/license_data.txt
```diff
@@ -0,0 +1,77 @@
...
+    "name": 'All Rights Reserved'
+  },
+  'http://creativecommons.org/publicdomain/mark/1.0/': {
```
sharpattack, line 5: Just curious, how were you able to map this url to their name and licenseId? At a glance, it seems to me the map on SPDX is removing the '/legalcode' part of the "seeAlso"/Other webpages.

sharpattack (approved):
Great job. I really like your solutions to provide a `_data.txt` as a way to create these mappings. I also appreciate the extra effort you give to make the data files.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -104,6 +105,8 @@
           "languages": ["eng"],
           "sizes": ["#{generic_file.page_count} pages"],
           "formats": "application/pdf",
+          "rights": [{"rights": 'Creative Commons Attribution Non Commercial Share Alike 3.0 United States', "scheme": "spdx", "identifier": \
+                      "CC-BY-NC-SA-3.0-US", "url": "http://creativecommons.org/licenses/by-nc-sa/3.0/us/"}],
```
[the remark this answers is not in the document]
Meowcenary: It's a super useful function. Got to love the enumerator module.

app/models/concerns/galtersufia/generic_file/license_data.txt
Meowcenary: This was one I actually had some questions on still. As I understand it "All Rights Reserved" isn't really a license in the sense that it doesn't give ANY rights to anyone else to use it. It's basically a statement saying you can't do anything with this. That said, we can probably just add an identifier that we all agree on. I'm not sure what it would look like though. Maybe since the rest of the identifiers are based on acronyms we could use something like `ARR`?

app/models/concerns/galtersufia/generic_file/license_data.txt
Meowcenary: It was really tedious, but I just went through the SPDX JSON files and searched for each license that was included in the Sufia initializer/configuration. The licenses were all mapped within the initializer as the name of the license to the URL fo the website of that license. So, basically I just took the URLs from the initializer as a key and then either copied the relevant JSON from the files I downloaded or just added the license id and name for the licenses that were not included on the SPDX.

Meowcenary: Alright, updated.

Issue #834, opened 2021-02-26
Export Metadata Subfields Related Identifiers (aka Related Works)


fenekku:
Export related identifiers / related work: per https://inveniordm.docs.cern.ch/reference/metadata/#related-identifiers-0-n
The recent changes in resource_type affects this too.

**Edit 2021-06-16**
As of writing, the current implementation accepts this structure:

```json
{
  "related_identifiers": [{
      "identifier": "10.1234/foo.bar",
      "scheme": "doi",
      "relation_type": "cites",
      "resource_type": {"id": "dataset"}
  }, ...],
}
```

Notice that **relation_type** should be used and not just **relation**.

However, it is likely that in the future (per this issue https://github.com/inveniosoftware/invenio-rdm-records/issues/635), the structure will be:

```json
{
  "related_identifiers": [{
      "identifier": "10.1234/foo.bar",
      "scheme": "doi",
      "relation_type": {"id": "cites"},
      "resource_type": {"id": "dataset"}
  }, ...],
}
```

So let's go with that structure. On my side, the import can adapt to whatever the reality is. 
The identifier and scheme part is ok for now.

sharpattack:
We will use related_url. 

Couple Questions: 
- What is the scheme? Is it limited to the following list? 
  - ISBN10, ISBN13, ISSN, ISTC, DOI, Handle, EAN8, EAN13, ISNI ORCID, ARK, PURL, LSID, URN, Bibcode, arXiv, PubMed ID, PubMed Central ID, GND, SRA, BioProject, BioSample, Ensembl, UniProt, RefSeq, Genome Assembly.
- What is the relation? Can it be some generic text? 
- What is the resource_type? This is not required.

sharpattack:
Example record with a related_url, but possibly not on the scheme list. 
https://digitalhub.northwestern.edu/files/j3860692c

fenekku:
Answers:
- Schemes are indeed picked from that list. Anything from the deposit page dropdown is included. `PURL` / `URN` in that list implies `URL` but it should indeed be official. As of writing, the supported list is:

```python
    SCHEMES = [
        "ark",
        "arxiv",
        "bibcode",
        "doi",
        "ean13",
        "eissn",
        "handle",
        "igsn",
        "isbn",
        "issn",
        "istc",
        "lissn",
        "lsid",
        "pmid",
        "purl",
        "upc",
        "url",
        "urn",
        "w3id"
    ]
```

[...]

PR #852, opened 2021-03-24
[#799] Metadata Subfield Identifiers

### LINKED ISSUE #799: Export Metadata Subfields Identifiers
Export the Metadata subfield [Identifiers](https://inveniordm.docs.cern.ch/reference/metadata/#identifiers-0-n)

sharpattack:
Use doi for the metadata identifier. closes #799

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -54,7 +54,7 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
 
   def record_for_export(generic_file)
     {
-      "pids": invennio_pids(generic_file.doi.shift),
+      "pids": invenio_pids(generic_file.doi.shift),
       "metadata": invenio_metadata(generic_file),
       "provenance": invenio_provenance(generic_file.proxy_depositor, generic_file.on_behalf_of)
     }
@@ -92,7 +92,7 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
     Sufia::Export::VersionGraphConverter.new(gf.content.versions).versions
   end
 
-  def invennio_pids(doi)
+  def invenio_pids(doi)
     {
       "doi": {
         "identifier": doi, # doi is stored in an array
@@ -127,7 +127,7 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
       "contributors": contributors(gf.contributor),
       "dates": gf.date_created.map{ |date| {"date": date, "type": "other", "description": "When the item was originally created."} },
       "languages": gf.language.any?{ |lang| lang.downcase == ENGLISH} ? ["eng"] : "",
-      "identifiers": identifiers(gf.doi.shift),
+      "identifiers": ark_identifiers(gf.ark),
       "sizes": Array.new.tap{ |size_json| size_json << "#{gf.page_count} pages" if !gf.page_count.blank? },
       "formats": gf.mime_type,
       "rights": rights(gf.rights),
@@ -248,11 +248,13 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
     end
   end
 
-  def identifiers(doi)
-    [{
-      "identifier": doi,
-      "scheme": "doi"
-    }]
+  def ark_identifiers(arks)
+    arks.map do |ark|
+      {
+        "identifier": ark,
+        "scheme": "ark"
+      }
+    end
   end
 
   def rights(license_urls)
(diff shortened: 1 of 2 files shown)


sharpattack:
Will need to revisit next month when details of the identifier metadata subfield has been ironed out.

carsonicator (approved):
Thanks @sharpattack. There will be a similar PR for the related identifier (#834), right?

app/models/invenio_rdm_record_converter.rb
```diff
@@ -127,6 +127,7 @@ def invenio_metadata(gf)
       "contributors": contributors(gf.contributor),
       "dates": gf.date_created.map{ |date| {"date": date, "type": "other", "description": "When the item was originally created."} },
       "languages": gf.language.any?{ |lang| lang.downcase == ENGLISH} ? ["eng"] : "",
+      "identifiers": identifiers(gf.doi.shift),
```
fenekku, line 130: Ah! Sorry I thought we were talking about the pids (persistent identifiers) where a managed DOI is stored. That was done above (btw there is a typo in that function name `invennio_pids` -> `invenio_pids`). For this `"identifiers"`, it is pure metadata. That means placing ARK identifiers here would be the right place to do so instead of the DOI. Any alternate identifiers go in this "identifiers" field.

If there are records with multiple DOIs, then the other ones could be placed in "identifiers" too. InvenioRDM only manages one DOI per record. On that note, I noticed the use of [shift](https://ruby-doc.org/core-2.5.0/Array.html#method-i-shift) which seems to mutate the array according to example (I thought `!` was used in Ruby to denote mutation?). How are the tests passing if `invennio_pids` shifts and then identifiers shifts again? Perhaps it is a different `shift` method?

fenekku: > As for alternate identifiers, we can add them here, but what goes into "related identifiers" metadata then?

This is a grey area so I don't think we can go wrong as long as we are consistent. Essentially any identifier that resolves to this record page should go into this `"identifiers"`. Any identifier that is complementary (e.g. has a relationship type like "is related to") or resolves to a record on a different platform should go in `"related_identifiers"`(e.g., pudmed ids and such).  
So we pick a strategy and keep consistent with it and then librarians can tell us if it makes sense to them.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -54,7 +54,7 @@ def generic_file_content_path(checksum)
 
   def record_for_export(generic_file)
     {
-      "pids": invennio_pids(generic_file.doi.shift),
+      "pids": invenio_pids(generic_file.doi.shift),
```
Meowcenary, line 57: Did we ever figure out what was going on with shift here? I looked at this a little bit thinking that maybe the generic_file argument was a copy of the generic_file, not a reference, and that would mean the mutation was only happening to a copy. I didn't actually confirm this though.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -92,7 +92,7 @@ def versions(gf)
     Sufia::Export::VersionGraphConverter.new(gf.content.versions).versions
   end
 
-  def invennio_pids(doi)
+  def invenio_pids(doi)
```
Meowcenary, line 95: I'm glad you did this. After @fenekku pointed it out I was sad.

Meowcenary (approved):
Looks good to me. Thanks for the changes!

app/models/invenio_rdm_record_converter.rb
sharpattack: Yes, I looked into this. It's actually because `generic_file` here is an ActiveFedora object. Calling `shift` won't mutate because it does not trigger it's datastore update.

app/models/invenio_rdm_record_converter.rb
sharpattack: 😆 same here, I reviewed it.

app/models/invenio_rdm_record_converter.rb
fenekku: Curious about that shift not updating the datastore: 
- would the changes still be in memory though?
- would make sense to me that they are not if: 
  * `generic_file.doi` is a join everytime (although I doubt it) OR
  * `generic_file` here is not the same in-memory `gf` as below. In other words there is 2 DB fetches of the same file OR
  * that shift has been overridden

Just curious :thinking: .

sharpattack: It is very misleading/confusing. I am glad you caught that `shift` should mutate under normal circumstances. Otherwise I may have continued to use in other cases with unintended consequences. 

This is what I know about Active**Fedora** and Active**Record**: is that they don't have the same interface, with ActiveFedora being much more limited. I have had trouble in the past with things like using `<<` to append to what should be an Array, but when trying to commit/save, it didn't. So, my guess is that `shift` returned the value first, then tried to remove the item from the "Array"(the commit/save part), but exhibited the same behavior as when I tried using the shovel operator. That's what I meant by `shift` not triggering the update, but should have said failed to make the update. So, it must be that `shift` is overriden.

sharpattack:
Closing without merging. 

There are still details that need to be worked out regarding what information to pull from DigitalHub's records and placed into either "identifiers" or "related identifiers".

app/models/invenio_rdm_record_converter.rb
fenekku: Really cool to know. Thanks for the explanation @Meowcenary.

PR #858, opened 2021-03-31
[#802] Export Metadata Subfield Versions

### LINKED ISSUE #802: Export Metadata Subfield Version
Export the Metadata subfield [Version](https://inveniordm.docs.cern.ch/reference/metadata/#version-0-1). 

Take particular care with this field as there is code in the [sufia export directory  ](https://github.com/samvera-deprecated/sufia/tree/6.x-stable/lib/sufia/export) that seems related to this.

Meowcenary:
closes #802

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -10,7 +10,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   SUBJECT_SCHEMES = [:tag, :mesh, :lcsh]
   ENG = "eng"
   ENGLISH = "english"
-  MEMOIZED_PERSON_OR_ORG_DATA_FILE = 'memoized_person_or_org_data.txt'
   ROLE_OTHER = 'other'
   DEFAULT_RIGHTS_SCHEME = "spdx"
   MEMOIZED_PERSON_OR_ORG_DATA_FILE = 'memoized_person_or_org_data.txt'
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -87,11 +87,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
     # @permissions = permissions(generic_file)
   end
 
-  def versions(gf)
-    return [] unless gf.content.has_versions?
-    Sufia::Export::VersionGraphConverter.new(gf.content.versions).versions
-  end
-
   def invennio_pids(doi)
     {
       "doi": {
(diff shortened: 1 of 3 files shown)


spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -131,7 +132,7 @@
       },
       "file": {
         "filename": generic_file.filename,
-        "content_path": nil # there is no file/content attached with this factory made GewnericFile
+        "content_path": nil
```
Meowcenary, line 135: I had removed this content while working on something that _did_ change the content, but I can re-add it. I forgot that it had been removed.

spec/support/sufia_factories.rb
```diff
@@ -1,11 +1,23 @@
-def make_generic_file(user, args = {})
+def make_generic_file(user, args={})
+  gf = build_generic_file(user, args)
+  gf.save!
+  gf
+end
+
+def make_generic_file_with_content(user, args={})
```
Meowcenary, line 7: I was running into two issues while testing so I ended up not using this, but this function will generate a GenericFile that has content attached to it. Because it has content it will also have versions.

Here are the two issues I'm running into:

1. Adding content to the let variable generic_file causes to_json to have a content path which will not be consistent across environments.
2. For some reason generating a second instance of InvenioRdmRecordConverter that uses a file with content causes duplicate constant errors which I think are in turn causing more errors. It was a really noisy error message, but the root of it seemed to be with constants being defined twice.

Meowcenary:
Tried to add some more thorough testing, but it turned into a mess. I left some comments on the code if you're interested in trying to get it working for yourself, but otherwise it's a simple PR.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -87,11 +86,6 @@ def record_for_export(generic_file)
     # @permissions = permissions(generic_file)
   end
 
-  def versions(gf)
-    return [] unless gf.content.has_versions?
-    Sufia::Export::VersionGraphConverter.new(gf.content.versions).versions
-  end
```
sharpattack, line 93: Thanks for removing this.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -263,4 +258,11 @@ def rights(license_urls)
       }
     end
   end
+
+  def version(content)
+    return "" unless content.has_versions?
+    version_number = content.versions.all.length
+
+    "v#{version_number}.0.0"
```
sharpattack, line 266: Simply replicating the documentation's example? When would a GenericFile's versions be more than one? Just curious, and we can omit testing this scenario.

spec/support/sufia_factories.rb
sharpattack: Thanks for keeping this in the commit. I ended up using it very minimally, and as long as there is some version, the `let(:json)` object is good enough for me.

sharpattack (approved):
Looks good to me.

PR #859, opened 2021-04-07
[#803] Export Metadata Subfield Rights All Rights Reserved

Meowcenary:
Update InvenioRdmRecordConverter#rights to export expected json for all
rights reserved license.

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -15,6 +15,7 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   INVENIO_PUBLIC = "public"
   INVENIO_RESTRICTED = "restricted"
   DEFAULT_RIGHTS_SCHEME = "spdx"
+  ALL_RIGHTS_RESERVED = 'All rights reserved'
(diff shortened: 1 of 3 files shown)


Meowcenary:
There were bugs with the rights field during export on staging and production. This should resolve any issues with licenses/rights.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -263,12 +264,18 @@ def rights(license_urls)
...
+      if license_data.present?
+        {
+          "rights": license_data[:"name"],
+          "scheme": DEFAULT_RIGHTS_SCHEME,
+          "identifier": license_data[:"licenseId"],
+          "url": license_url
+        }
+      elsif license_url == ALL_RIGHTS_RESERVED
+        {
+          "rights": ALL_RIGHTS_RESERVED
```
sharpattack, line 276: How about using `ALL_RIGHTS_RESERVED` with `@@license_data`? Since we have that data, it would fill in the other fields(scheme, identifier, and url), right?

sharpattack (approved):
Looks good to me. I am fine with just setting`rights` to default to "All rights reserved", but want to make sure it doesn't matter that we omit the other fields. Also makes me wonder what other rights were there that broke this?

PR #862, opened 2021-04-12
[#861] Special Characteres Break Header Queries

### LINKED ISSUE #861: Special Characters Break Header Queries
The DigitalHub export is breaking when mesh queries with ampersands are submitted. Fix mesh queries for special characters and then check to see if this is an issue for LCSH queries as well.

Meowcenary:
Encode mesh and lcsh searches to support special characters. closes #861

### DIFF
--- a/app/models/header_lookup.rb
+++ b/app/models/header_lookup.rb
@@ -28,7 +28,7 @@ class HeaderLookup
 
   # return PID for provided mesh_header using SPARQL query
   def mesh_term_pid_lookup(mesh_term="")
-    hits = perform_and_parse_mesh_query(mesh_term)
+    hits = perform_and_parse_mesh_query(CGI.escape(mesh_term))
 
     if hits.present?
       mesh_pid = pid_from_mesh_hits(hits)
@@ -43,7 +43,7 @@ class HeaderLookup
   # lookup lcsh term, memoize it, write it to file, return PID
   def lcsh_term_pid_lookup(lcsh_term="")
     stripped_lcsh_term = strip_accents(lcsh_term)
-    subject_names, subject_id_uris = perform_and_parse_lcsh_query(stripped_lcsh_term)
+    subject_names, subject_id_uris = perform_and_parse_lcsh_query(CGI.escape(stripped_lcsh_term))
 
     if subject_names.present? && subject_id_uris.present?
       lcsh_pid = pid_from_lcsh_hits(lcsh_term, stripped_lcsh_term, subject_names, subject_id_uris)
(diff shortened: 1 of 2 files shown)


spec/models/header_lookup_spec.rb
```diff
@@ -7,8 +7,8 @@
     ".w3.org%2F2000%2F01%2Frdf-schema%23%3E%0D%0APREFIX%20meshv%3A%20%3Chttp%3A%2F%2Fid.nlm.nih.gov%2Fmesh%2Fvocab%23%3E"\
     "%0D%0APREFIX%20mesh2018%3A%20%3Chttp%3A%2F%2Fid.nlm.nih.gov%2Fmesh%3E%0D%0A%0D%0ASELECT%20%3Fd%20%3FdName%0D%0AFROM"\
     "%20%3Chttp%3A%2F%2Fid.nlm.nih.gov%2Fmesh%3E%0D%0AWHERE%20%7B%0D%0A%20%20%3Fd%20a%20meshv%3ADescriptor%20.%0D%0A%20%"\
-    "20%3Fd%20rdfs%3Alabel%20%3FdName%0D%0A%20%20FILTER(REGEX(%3FdName%2C%27Vocabulary, Controlled%27%2C%20%27i%27))%20%0"\
-    "D%0A%7D%20%0D%0AORDER%20BY%20%3Fd%20%0D%0A"
+    "20%3Fd%20rdfs%3Alabel%20%3FdName%0D%0A%20%20FILTER(REGEX(%3FdName%2C%27Vocabulary%2C+Controlled%27%2C%20%27i%27))%2"\
```
Meowcenary, line 10: The urls are different because the headers are encoded now.

spec/models/header_lookup_spec.rb
```diff
@@ -70,7 +70,7 @@
   end
 
   let(:lcsh_term) { "Semantic Web" }
-  let(:lcsh_query_url) { "http://id.loc.gov/authorities/subjects/suggest/?q=*Semantic*Web*" }
+  let(:lcsh_query_url) { "http://id.loc.gov/authorities/subjects/suggest/?q=*Semantic+Web*" }
```
Meowcenary, line 73: same deal as above, but less of change

app/models/header_lookup.rb
```diff
@@ -28,7 +28,7 @@ def pid_lookup_by_scheme(term="", scheme="")
 
   # return PID for provided mesh_header using SPARQL query
   def mesh_term_pid_lookup(mesh_term="")
-    hits = perform_and_parse_mesh_query(mesh_term)
+    hits = perform_and_parse_mesh_query(CGI.escape(mesh_term))
```
Meowcenary, line 31: This is the fix itself. `Obesity-- prevention & control` does not actually return anything, but at least now it can be searched without breaking things.

app/models/header_lookup.rb
```diff
@@ -43,7 +43,7 @@ def mesh_term_pid_lookup(mesh_term="")
   # lookup lcsh term, memoize it, write it to file, return PID
   def lcsh_term_pid_lookup(lcsh_term="")
     stripped_lcsh_term = strip_accents(lcsh_term)
-    subject_names, subject_id_uris = perform_and_parse_lcsh_query(stripped_lcsh_term)
+    subject_names, subject_id_uris = perform_and_parse_lcsh_query(CGI.escape(stripped_lcsh_term))
```
Meowcenary, line 46: I think that the same string, `Obesity-- prevention & control` appears in the LCSH terms as well so I figured it made sense to change it here as well.

sharpattack (approved):
Thank you for fixing.

PR #863, opened 2021-04-14
Release 1.21.03 Invenio Export Support Special Characters In Headers

Meowcenary:
Fix bug with special characters breaking header lookup.


sharpattack (approved):
Thanks for releasing. 🤞 the export runs without error this time.

PR #871, opened 2021-06-03
[#864] Nil Value Breaking Build Creator Contributor JSON

### LINKED ISSUE #864: Nil Value Breaking Build Creator Contributor Json
`InvenioRdmRecordConverter#build_creator_contributor_json` is returning nil values when splitting creator name. Check what record is causing this on DigitalHub production and put in code to fix the bug.

Meowcenary:
Add rake tasks to help with debugging. Log file id and title while
running export. closes #864

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,3 +74,13 @@ digital-repository/$ rspec spec/
 ```
 digital-repository/$ cap <stage> deploy
 ```
+
+##### Digital Hub Export
+* Export repo to JSON formatted for InvenioRDM
+```
+./bin/rake repo_export
+```
+* Same as above, but with extra debugging information
+```
+./bin/rake verbose debug repo_export
+```
--- a/lib/tasks/debug.rake
+++ b/lib/tasks/debug.rake
@@ -1,0 +1,14 @@
+desc "switch rails logger to stdout"
+task :verbose => [:environment] do
+  Rails.logger = Logger.new(STDOUT)
+end
+
+desc "switch rails logger log level to debug"
+task :debug => [:environment, :verbose] do
+  Rails.logger.level = Logger::DEBUG
+end
+
+desc "switch rails logger log level to info"
+task :info => [:environment, :verbose] do
+  Rails.logger.level = Logger::INFO
+end


README.md
```diff
@@ -74,3 +74,13 @@ digital-repository/$ rspec spec/
...
 ```
+
+##### Digital Hub Export
+* Export repo to JSON formatted for InvenioRDM
+```
+./bin/rake repo_export
+```
+* Same as above, but with extra debugging information
+```
+./bin/rake verbose debug repo_export
```
Meowcenary, line 85: It's weird, but chaining the rake tasks persists the logger and it's config.

Meowcenary:
The original ticket was to resolve an error that was breaking the export, but from what I was seeing the export was running without issue. I thought it would still be useful to merge what I was using to help with debugging.

lib/tasks/debug.rake
```diff
@@ -0,0 +1,14 @@
...
+
+desc "switch rails logger log level to debug"
+task :debug => [:environment, :verbose] do
+  Rails.logger.level = Logger::DEBUG
+end
+
+desc "switch rails logger log level to info"
+task :info => [:environment, :verbose] do
+  Rails.logger.level = Logger::INFO
+end
```
sharpattack, line 14: This is a nice feature, and I plan to use it.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -31,6 +31,8 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   def initialize(generic_file=nil)
     return unless generic_file
 
+    $stdout.puts "processing id: #{generic_file.id}, title: #{generic_file.title.first}"
```
sharpattack, line 34: Will this get logged to the screen, even without the `verbose` and/or `debug` options for the rake task? And with the rake task options, it will be logged to a log file? 

What about limiting logging to only the `generic_file` ones that break?

README.md
sharpattack: I see, I was wondering why both "verbose" and "debug" were needed.

app/models/invenio_rdm_record_converter.rb
Meowcenary: I'm going to check this to make sure. I was having trouble getting output to the screen without the options, but that might have been something wrong with the code itself.

Meowcenary: It takes a really, really long time for it to start actually processing files. It always throws me off.

sharpattack (approved):
Look good, thanks for adding the debugging tool!

PR #878, opened 2021-06-16
[#856] Revisit Licenses/Rights

### LINKED ISSUE #856: Revisit Metadata Subfield Licenses
The next release of InvenioRDM will have updates to the metadata subfield licenses. Check for any updates that need to be made and implement those changes.

**EDIT 2021-05-28**

Licenses ("rights" field) should be like this:

```json
{
  "record": {
    "metadata": {
      "rights": [{
        "id": "CC-BY-3.0-US",
        "link": "http://creativecommons.org/licenses/by/3.0/us/",
        "title": "Creative Commons Attribution 3.0 United States"
      },
      ...
    }
  }
}
```

modified to appropriate licenses. Valid `id`s can be found here: https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/fixtures/data/vocabularies/licenses.csv

sharpattack:
Update licenses/rights to new format. closes #856

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -14,7 +14,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   OPEN_ACCESS = "open"
   INVENIO_PUBLIC = "public"
   INVENIO_RESTRICTED = "restricted"
-  DEFAULT_RIGHTS_SCHEME = "spdx"
   ALL_RIGHTS_RESERVED = 'All rights reserved'
   MEMOIZED_PERSON_OR_ORG_DATA_FILE = 'memoized_person_or_org_data.txt'
   FUNDING_DATA_FILE = 'app/models/concerns/galtersufia/generic_file/funding_data.txt'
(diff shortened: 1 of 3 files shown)


Meowcenary:
I'll take a look at this in a moment.

Meowcenary (approved):
Thanks for handling this!

app/models/invenio_rdm_record_converter.rb
```diff
@@ -267,14 +266,13 @@ def rights(license_urls)
 
       if license_url == ALL_RIGHTS_RESERVED
         {
-          "rights": license_data[:"name"],
+          "title": license_data[:"name"]
```
fenekku, line 269: Since `id` is required, but it's hard to pick an `id` that is unique, what about adding

```
"id": "GALTER-ARR-1" 
```
It doesn't conflict with anything on the SPDX list and the `GALTER` prefix namespaces it enough. The `1` is just to futureproof against legalese change.

sharpattack: Okay, did not realize it was required. I like the suggested 'id', and will add it.

sharpattack: Oh what about link? Is that required?

sharpattack: Never mind, I see that link is not required. Unless the documentation is not updated, just let me know and we can have a quick chat with the group about this.

fenekku: The docs are good on that one :sweat_smile: : `link` is not required. The best would be to not provide the key either in that case.

sharpattack: @fenekku can I take your above comment as an approval of this PR then? `link` is omitted from "All Rights Reserved". No other case may have an "empty" link.

fenekku (approved):
Yes that works. Thanks for that.

PR #879, opened 2021-06-16
799 subfield identifiers

sharpattack:
Re-opened from earlier PR.

No changes since that PR, other than this has the latest changes.

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -140,6 +140,7 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
       "contributors": contributors(gf.contributor),
       "dates": gf.date_created.map{ |date| {"date": date, "type": "other", "description": "When the item was originally created."} },
       "languages": gf.language.map{ |lang| lang.present? && lang.downcase == ENGLISH ? {"id": "eng"} : nil }.compact,
+      "identifiers": identifiers(gf.doi.shift),
       "related_identifiers": related_identifiers(gf.related_url),
       "sizes": Array.new.tap{ |size_json| size_json << "#{gf.page_count} pages" if !gf.page_count.blank? },
       "formats": gf.mime_type,
@@ -161,14 +162,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
       {
         "id": "other-other"
       }
-    end
-  end
-
-  def contributors(contributors)
-    contributors.map do |contributor|
-      contributor_json = build_creator_contributor_json(contributor)
-      contributor_json[:person_or_org].merge!({"role": ROLE_OTHER})
-      contributor_json
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/models/invenio_rdm_record_converter.rb
```diff
@@ -57,7 +57,7 @@ def generic_file_content_path(checksum)
 
   def record_for_export(generic_file)
     {
-      "pids": invennio_pids(generic_file.doi.shift),
+      "pids": invenio_pids(generic_file.doi.shift),
```
Meowcenary, line 60: oof, good catch!

app/models/invenio_rdm_record_converter.rb
```diff
@@ -226,6 +219,18 @@ def build_creator_contributor_json(creator)
     File.write(MEMOIZED_PERSON_OR_ORG_DATA_FILE, @@person_or_org_data)
     # return the actual json
     json
+  end # build_creator_contributor_json
+
+  def additional_titles(titles)
```
Meowcenary, line 224: This function is fine as it is, but here's another take for you to consider. I'll leave it up to you if you'd like to change anything or not. 

```
def additional_titles(additional_titles)
  return nil if additional_titles.blank?
  additional_titles.map{ |title| {"title": title, "type": "alternative_title", "lang": ENG} }
end

additional_title(titles.shift)
```

app/models/invenio_rdm_record_converter.rb
```diff
@@ -226,6 +219,18 @@ def build_creator_contributor_json(creator)
...
+    additional_titles_size = titles.size-1
+    return nil if additional_titles_size < 0
+    titles.last(additional_titles_size).map{ |title| {"title": title, "type": "alternative_title", "lang": ENG} }
+  end
+
+  def additional_descriptions(descriptions)
```
Meowcenary, line 230: Same as the above comment. The block in that comment can be changed to fit here pretty easily. Actually, if you wanted to abstract a bit you could even write one function that handles both of these in one, but it would probably be more trouble than it's worth.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -245,20 +250,21 @@ def subjects_for_scheme(terms, scheme)
     end
   end
 
-  def additional_titles(titles)
```
Meowcenary, line 248: haha, seems like I was the one who wrote this in the first place. I guess on a second glance I see things to change.

Meowcenary (approved):
Looks good to me. Some small comments, but nothing that I think necessarily needs to change.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -357,12 +361,29 @@
...
+      "identifier": "MIT",
+      "url": mit_license_url
+    },
+    {
+      "rights": "Creative Commons Zero v1.0 Universal",
+      "scheme": "spdx",
+      "identifier": "CC0-1.0",
+      "url": creative_commons_zero_url
+    }]
+  end
```
fenekku, line 378: Are these the expected exported rights? If so, they will be changed for https://inveniordm.docs.cern.ch/reference/metadata/#rights-0-n correct?

sharpattack: Ah, yes. This will need to be updated once #878 is merged.

app/models/invenio_rdm_record_converter.rb
sharpattack: 😄 It happens to us all. I do like your suggested changes, first giving a shot at abstracting it.

Meowcenary:
I just merged a PR that seems to have caused conflicts. If you have any trouble resolving these let me know.

fenekku (approved):

app/models/invenio_rdm_record_converter.rb
sharpattack: There were more bugs uncovered while trying to abstract. Opened a new issue, #887 to address it separately.

PR #880, opened 2021-06-16
[#27] Update Exported Languages

### LINKED ISSUE #27: Large file upload takes a long time and doesn't take you to the next screen of assigning descriptive metadata

Meowcenary:
Update structure for Invenio export "languages" field to match release
for June 2021. closes #27

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -135,7 +135,7 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
       "subjects": SUBJECT_SCHEMES.map{ |subject_type| subjects_for_scheme(gf.send(subject_type), subject_type) }.flatten,
       "contributors": contributors(gf.contributor),
       "dates": gf.date_created.map{ |date| {"date": date, "type": "other", "description": "When the item was originally created."} },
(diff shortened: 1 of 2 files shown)


app/models/invenio_rdm_record_converter.rb
```diff
@@ -135,7 +135,7 @@ def invenio_metadata(gf)
       "subjects": SUBJECT_SCHEMES.map{ |subject_type| subjects_for_scheme(gf.send(subject_type), subject_type) }.flatten,
       "contributors": contributors(gf.contributor),
       "dates": gf.date_created.map{ |date| {"date": date, "type": "other", "description": "When the item was originally created."} },
-      "languages": gf.language.any?{ |lang| lang.downcase == ENGLISH} ? ["eng"] : "",
+      "languages": gf.language.map{ |lang| lang.present? && lang.downcase == ENGLISH ? {"id": "eng"} : nil }.compact,
```
fenekku, line 138: Minor feel free to ignore if too much effort:
- InvenioRDM does accept other languages than "eng". The mapping here might be more complicated than it's worth though. 
- If a field is empty, it doesn't have to be listed. For instance, if there are no languages, `"languages"` doesn't have to be included in the final json. This holds for any optional field and may save us some space when all done (e.g. no "sizes", then no need to include it in final json). This is not required `"languages": []` will parse fine. This may be more effort though.

fenekku (approved):

PR #881, opened 2021-06-17
834 export metadata subfield related works

sharpattack:


app/models/invenio_rdm_record_converter.rb
```diff
@@ -286,4 +288,32 @@ def version(content)
 
     "v#{version_number}.0.0"
   end
+
+  def related_identifiers(related_url)
```
Meowcenary, line 292: This is a good function. Very clear.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -286,4 +288,32 @@ def version(content)
...
+          "scheme": "url",
+          "relation_type": {"id": "isRelatedTo"}
+        }
+      end
+    end
+
+    identifiers.compact
+  end
+
+  def doi_url?(url)
```
Meowcenary, line 316: Nice!

Meowcenary (approved):
Looks good to me. Nice job with this.

app/models/invenio_rdm_record_converter.rb
sharpattack: This is your work actually 😄

app/models/invenio_rdm_record_converter.rb
sharpattack: Same here, 😎

PR #882, opened 2021-06-17
[#54] Add Resource Types YAML

### LINKED ISSUE #54: What to do with keywords
Currently the GVB collection doesn't have any keywords and thus it doesn't feature in the cloud on the front page. Should we treat subjects as keywords?

Meowcenary:
Add resource types YAML file. closes https://github.com/galterlibrary/digitalhub-2/issues/54


Meowcenary:
If you're curious you can compare against this file: https://github.com/galterlibrary/digitalhub-2/blob/201359e4d685503c4cd66933c6e04b699adf0cc0/app_data/vocabularies/resource_types.yaml . What I did to check my work was pick random types from the YAML and then search for them in the linked file. Obviously everything you search should match to an ID in that file otherwise there's a problem.

fenekku (commented):
This file only contains the subtype is that on purpose? How is the code using this file?

sharpattack:
Sorry @Meowcenary, I should have been clearer. This task was not meant for this repository, but for galterlibrary/digitalhub-2. It looks like @fenekku has already done this, but is not merged into main, but is on the may_release branch. Unless I missed something else, I don't think we need this file in this code base.

Meowcenary:
Ah, okay. I'll close the PR. 

@fenekku , the intention was to load the YAML file and then use the hash to find the ID for each resource type. I was under the impression that we were changing how we handled the export by using this instead of the previous mapping, but I guess that's not right.

PR #884, opened 2021-06-21
[#26] Update Resource Types Field

### LINKED ISSUE #26: Multiple files upload takes long and looks stuck

Meowcenary:
Update resource types field to latest version of InvenioRDM. closes #26

### DIFF
--- a/app/models/concerns/galtersufia/generic_file/invenio_resource_type_mappings.rb
+++ b/app/models/concerns/galtersufia/generic_file/invenio_resource_type_mappings.rb
@@ -25,7 +25,7 @@ module Galtersufia
         "Collected Correspondence": ["archival", "archival-collection"],
         "Conference Proceeding": ["conference", "conference-conference_proceeding"],
         "Congresses": ["conference", "conference-congress"],
-        "Dataset": ["dataset"],
+        "Dataset": ["dataset", "dataset"],
         "Dictionary": ["book", "book-book"],
         "Drawings": ["image", "image-drawing"],
         "Editorial Article": ["article", "article-editorial"],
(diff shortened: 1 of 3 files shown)


app/models/concerns/galtersufia/generic_file/invenio_resource_type_mappings.rb
```diff
@@ -25,7 +25,7 @@ module InvenioResourceTypeMappings
         "Collected Correspondence": ["archival", "archival-collection"],
         "Conference Proceeding": ["conference", "conference-conference_proceeding"],
         "Congresses": ["conference", "conference-congress"],
-        "Dataset": ["dataset"],
+        "Dataset": ["dataset", "dataset"],
```
Meowcenary, line 28: This was the only mapping that did not have two values. Adding the extra value makes the logic a little simpler in the converter. 

I did not remove the first value, the parent type, because I figure we might want it for the future. I didn't consider this before now, but since it's stored in git we could remove it without it being a problem and have only the type id stored here. If that's something we would prefer I can quickly make the change.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -154,19 +154,13 @@ def invenio_metadata(gf)
   def resource_type(digitalhub_subtype)
     irdm_types = DH_IRDM_RESOURCE_TYPES[digitalhub_subtype]
 
-    if irdm_types && irdm_types[1]
+    if irdm_types
```
Meowcenary, line 157: Explanation of this change above.

app/models/concerns/galtersufia/generic_file/invenio_resource_type_mappings.rb
sharpattack: Great change! Much better to add this and remove the logic/branch below.

sharpattack (approved):
Looks good to me. Thank you for making the changes.

app/models/concerns/galtersufia/generic_file/invenio_resource_type_mappings.rb
Meowcenary: Thanks! Yeah, as we've gone back to update things I've found myself seriously wondering what was running through my head the first time lol.

PR #885, opened 2021-06-21
[#28] Update Exported Role

### LINKED ISSUE #28: Figure out what the batch functionality is and if we need it.
Is it different then 'Folder Upload'

Meowcenary:
Update export fields Role and Contributions for lastest InvenioRDM.
closes https://github.com/galterlibrary/galter_ir_exporter/issues/28


spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -60,14 +60,16 @@
...
-            }
+               "type": "personal",
+               "given_name": "#{user.formal_name.split(',').last}",
+               "family_name": "#{user.formal_name.split(',').first}",
+               "identifiers": {
+                 "scheme": "orcid",
+                 "identifier": "#{user.orcid.split('/').last}"
+               }
+            },
+            "role": {}, # TBD on June InvenioRDM release, will contain "id": <id of role>
```
Meowcenary, line 71: This could probably be bounced out into it's own `let(:)` variable, but I know we are going to change this in the near future when the role field is finalized for the June release. I figure this will do for now and we can DRY things out after we know what we're dealing with.

memoized_person_or_org_data.txt
```diff
@@ -1 +1 @@
...
+{"Tester, Mock"=>{:person_or_org=>{"type"=>"personal", "given_name"=>" Mock", "family_name"=>"Tester", "identifiers"=>{:scheme=>"orcid", :identifier=>"1234-5678-9123-4567"}}}, "User, Contributor"=>{:person_or_org=>{"type"=>"personal", "given_name"=>" Contributor", "family_name"=>"User"}}, "Laster, Firston"=>{:person_or_org=>{"type"=>"personal", "given_name"=>"Firston", "family_name"=>"Laster"}}, "Firston Laster"=>{:person_or_org=>{"name"=>"Firston Laster", "type"=>"organisational"}}, "Unknown"=>{:person_or_org=>{"name"=>"Unknown", "type"=>"organisational"}}, "Creator not identified."=>{:person_or_org=>{"name"=>"Creator not identified.", "type"=>"organisational"}}, "Galter Health Sciences Library"=>{:person_or_org=>{"name"=>"Galter Health Sciences Library", "type"=>"organisational"}}}
```
Meowcenary, line 1: It's hard to see, but because of the change in the structure the memoization needed to change as well or specs would just keep failing.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -174,14 +174,14 @@ def resource_type(digitalhub_subtype)
...
-      contributor_json[:person_or_org].merge!({"role": ROLE_OTHER})
+      contributor_json.merge!({"role": {}})
       contributor_json
     end
   end
 
   def build_creator_contributor_json(creator)
     if creator_data = @@person_or_org_data[creator]
-      return creator_data
+      return creator_data.merge({"role": {}, "affiliations": {}})
```
Meowcenary, line 184: Simple way to meet the requirements. Role is not determined yet and affiliations are all nil for production DigitalHub.

If it's not clear, I moved all the previous data we had here (except for role and affiliations) into the person_or_org field. So, really, all that's happened is that the previous information is now the value for the key "person_or_org" and the other two key-value pairs just hold empty hashes.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -186,7 +189,9 @@
         "type": "personal",
```
Meowcenary, line 189: This is a good place to see what I had thought the expected structure should be. Let me know if I got something wrong here, I definitely got a little confused myself with all the brackets and what not.

fenekku (commented):
I've added a first take on the roles vocabulary for Prism taken from the list the DWIG established (https://northwestern.app.box.com/file/725826115682) here: https://github.com/galterlibrary/digitalhub-2/blob/master/app_data/vocabularies/roles.yaml. The id should be taken from ids in that yaml file (and per usual "taken from" just means that valid ids are the ones in that file).

The question remaining is where the roles are taken from in digitalhub and what is their correspondence @sharpattack or @saragon02 might know.

Meowcenary:
> I've added a first take on the roles vocabulary for Prism taken from the list the DWIG established (https://northwestern.app.box.com/file/725826115682) here: https://github.com/galterlibrary/digitalhub-2/blob/master/app_data/vocabularies/roles.yaml. The id should be taken from ids in that yaml file (and per usual "taken from" just means that valid ids are the ones in that file).
> 
> The question remaining is where the roles are taken from in digitalhub and what is their correspondence @sharpattack or @saragon02 might know.

I'll bring this up during standup and we can go ahead and discuss on Slack or some other way if it gets complicated.

fenekku:
@Meowcenary Is this still valid? I haven't seen any roles in the export so perhaps it is. Do you need anything here?

Meowcenary:
@fenekku , I just read through this again and I think we can close it out. If I remember right we decided to not export this field.

PR #886, opened 2021-06-22
[#857] Revisit Metadata Subfield Subjects

### LINKED ISSUE #857: Revisit Metadata Subfield Subjects
The next release of InvenioRDM will have updates to the metadata subfield subjects. Check for any updates that need to be made and implement those changes.

**EDIT 2021-06-18**
The new subjects structure:

```json
{
    "subjects": [{
        "id": "D001921",
      }, 
      ...
    ],
}
```

For MeSH we will use the MeSH identifier (e.g. "D001921" as above). 
For FAST, we will use its identifier as a string (e.g. "862474")
For other subjects (if any) a proper identifier should be chosen.

All extra subject information like its human readable label will be loaded on the instance side.

Meowcenary:
Update metadata subfield subjects for June 2021 InvenioRDM release.
closes #857

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -7,11 +7,13 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   include Galtersufia::GenericFile::InvenioResourceTypeMappings
   include Galtersufia::GenericFile::KnownOrganizations
 
-  SUBJECT_SCHEMES = [:tag, :mesh, :lcsh]
+  SUBJECT_SCHEMES = [:tag, :mesh, :lcsh, :subject_name]
   CIRCA_VALUES = ["ca.", "ca", "circa", "CA.", "CA", "CIRCA"]
   UNDATED = ["undated", "UNDATED"]
+  REFERENCE_FIELDS = ["bibliographic_citation", "part_of"]
   ABBR_MONTHNAMES = Date::ABBR_MONTHNAMES.map{ |abbr_monthname| abbr_monthname.downcase if abbr_monthname.present? }
   MONTHNAMES = Date::MONTHNAMES.map{ |monthname| monthname.downcase if monthname.present? }
+  SEASONS = ["spring", "summer", "fall", "winter"]
   ENG = "eng"
   ENGLISH = "english"
   ROLE_OTHER = 'other'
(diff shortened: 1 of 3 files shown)


spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -532,4 +525,13 @@
       end
     end
   end
+
+  describe "#subjects_for_scheme" do
```
Meowcenary, line 529: Small unit test. It's going to be a pain to change things so that we do not include a key if it's blank, so for now I'm just returning values that would evaluate true for `#blank?`.

Meowcenary (commented):
@fenekku , I know you've mentioned on a few PRs that you would prefer that we do not include a key-value pair if the value will be blank. I think that's something we can definitely do, I've done similar things by passing a block to the Hash constructor: `Hash.new{ |h| do_stuff}`, but for now I don't want to mess with that too much since we're on a relatively tight schedule.

fenekku (approved):
It's fine if we don't get to remove the "subjects" key if empty. I thought @sharpattack was also working on this though. In particular the mapping of digitalhub subjects to Prism, no?

app/models/invenio_rdm_record_converter.rb
```diff
@@ -231,17 +231,19 @@ def build_creator_contributor_json(creator)
...
+          {id: pid}
         else
-          {subject: "#{term}: DigitalHub field #{scheme}"}
+          nil
         end
       end
+
+      mapped_terms.compact
     else
-      terms.map{ |term| {subject: term} }
```
sharpattack, line 244: I can't find a note of it, but I thought we were adding tags? I will check with our team to confirm.

sharpattack (commented):
@fenekku I thought I was too 😅 But I am glad Eric jumped in. I forgot he made a function that fetches the IDs. 

There are two other "subjects" I think we are not accounting for: "Name" and "Geographic Name". There is also "tag/keyword" I thought we were including here. I will check with the team on these.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -532,4 +525,13 @@
...
+    let(:unknown_mesh_term){ ["nothing but lies"] }
```
carsonicator, line 530: nice.

carsonicator (approved):
Thanks, Eric.

Meowcenary:
Alright, this has finally been squashed to a single commit, debugging stuff has been moved to a separate branch, and specs have been run. I'll mention it at standup in case anyone wants to review again quickly.

sharpattack (approved):
Thank you for cleaning up the branch!

PR #889, opened 2021-06-24
[#888] Generate Tags YAML File

### LINKED ISSUE #888: Generate Tags YAML File
Generate a YAML file with the following format:

```
- id: <a slug for the keyword>
  title:
    en: <human readable label>
```

The slug should be a value that is unique to each tag (easy way to do this is with `#parameterize`) and the human readable label should just be the raw keyword.

Meowcenary:
Script to generate YAML file of tags formatted for InvenioRDM.
closes #888

### DIFF
--- a/lib/scripts/create_keywords_yaml.rb
+++ b/lib/scripts/create_keywords_yaml.rb
@@ -1,0 +1,18 @@
+INDENT = "  "
+
+File.open("keywords.yaml", "w") do |file|
+  tags = {}
+
+  GenericFile.all.each do |gf|
+    gf.tag.each do |tag|
+      slug = tag.parameterize
+
+      if tags[slug]
+        next
+      else
+        tags[slug] = tag
+        file.write("- id: #{slug}\n#{INDENT}title:\n#{INDENT*2}en: \"#{tag}\"\n")
+      end
+    end
+  end
+end


lib/scripts/create_keywords_yaml.rb
```diff
@@ -0,0 +1,18 @@
+INDENT = "  "
+
+File.open("keywords.yaml", "w") do |file|
+  tags = {}
+
+  GenericFile.all.each do |gf|
+    gf.tag.each do |tag|
+      slug = tag.parameterize
```
Meowcenary, line 8: https://apidock.com/rails/ActiveSupport/Inflector/parameterize

fenekku: Sweet. `parametrize` handles commas, right?

lib/scripts/create_keywords_yaml.rb
```diff
@@ -0,0 +1,18 @@
...
+
+      if tags[slug]
+        next
+      else
+        tags[slug] = tag
+        file.write("- id: #{slug}\n#{INDENT}title:\n#{INDENT*2}en: #{tag}\n")
```
fenekku, line 14: I think I like this simple approach because it creates the yaml file iteratively. I have been using a yaml library on the Python side to write the yaml, but it takes the whole data as input which is not great. The only thing I wonder here is if the content of `#{tag}` could interfere with the yaml syntax... e.g. if it has colons, commas, dashes, brackets... Do our keywords look like that?

Feel free to make a PR to https://github.com/galterlibrary/digitalhub-2 with the generated keywords.yaml when you have it.

fenekku (approved):
Consider those remarks and if it's ok feel free to merge.

sharpattack (approved):
Thank you for adding this script to the codebase! Other than Guillaume's question about syntax interference, it looks good to me.

lib/scripts/create_keywords_yaml.rb
Meowcenary: I was thinking we could wrap it in quotes. I think that is valid YAML syntax.

lib/scripts/create_keywords_yaml.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  GenericFile.all.each do |gf|
+    gf.tag.each do |tag|
+      slug = tag.parameterize
+
+      if tags[slug]
+        next
+      else
+        tags[slug] = tag
+        file.write("- id: #{slug}\n#{INDENT}title:\n#{INDENT*2}en: \"#{tag}\"\n")
```
Meowcenary, line 14: Updated to use double quote character to wrap tag. This should handle any special characters that may appear in tags.

PR #890, opened 2021-06-25
887 additional refactor

sharpattack:


app/models/invenio_rdm_record_converter.rb
```diff
@@ -131,9 +131,9 @@ def invenio_metadata(gf)
       "resource_type": resource_type(gf.resource_type.shift),
       "creators": gf.creator.map{ |creator| build_creator_contributor_json(creator) },
       "title": gf.title.first,
-      "additional_titles": additional_titles(gf.title),
+      "additional_titles": additional(category: "title", array: gf.title),
```
sharpattack, line 134: I did not do the slicing of the array here only because of the tests. I would have to do the slicing of the arguments in the tests too. Knowing that, I thought you may have preferred the slicing to be done in the method instead. If not, just let me know and I can add your original suggestion back.

sharpattack: @fenekku ☝️

app/models/invenio_rdm_record_converter.rb
```diff
@@ -220,16 +220,23 @@ def build_creator_contributor_json(creator)
...
-  def additional_titles(titles)
-    additional_titles_size = titles.size-1
-    return nil if additional_titles_size < 0
-    titles.last(additional_titles_size).map{ |title| {"title": title, "type": "alternative_title", "lang": ENG} }
+  def additional(category:, array:)
+    type = set_type(category)
+
+    # return all values except the first, remove the strings that contain only
+    # spaces or are empty, then map
+    array.slice(1..-1).delete_if(&:blank?).map{ |word|
```
fenekku, line 228: Just curious: `array.slice(1..-1)` versus `array[1..]` here? I think it's because of wanting the delete to run on a copy of the original... What about using [`filter_map`](https://ruby-doc.org/core-2.7.0/Enumerable.html#filter_map-method) instead?

fenekku (approved):
That's good by me. Slicing before or inside doesn't change much so either way is fine. Minor note about the filtering process, but you can merge even if you keep things the same.

Meowcenary:
Ugh, I was looking at the open PRs and accidentally merged this and not my own PR. @sharpattack is out right now so I'll make a note to follow up with him on this.

app/models/invenio_rdm_record_converter.rb
sharpattack: Actually, no difference 😅, but you do have to specify the end. So `array[1..-1]` would have worked. 
Wish I knew to use `filter_map`, it will work nicely. Maybe I'll add it later.

fenekku: ah maybe I had a different version of Ruby:
```
> irb
irb(main):001:0> a = [1,2,3]
irb(main):002:0> a[1..]
=> [2, 3]
```

```
> irb --version
irb 1.2.1 (2019-12-24)
> ruby --version
ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-linux-gnu]
```

PR #894, opened 2021-07-16
[#892] extra_data presentation_location

### LINKED ISSUE #892: Add presentation_location to "extras" section
Not currently in the official documentation, but we will populate this field with DH `location`

sharpattack:
Add new `extra_data` field for export, populate with
"location/based_near" data as `presenation_location`. closes #892


fenekku (approved):
Interesting that it's called "based_near" rather than "location" or something like this.

sharpattack:
Agreed. This is not the only time something has been called one thing in the backend, and another on the frontend. Thank you for the quick review!

PR #897, opened 2021-07-21
[#891] Revisit Locations

### LINKED ISSUE #891: Revisit Locations Metadata Subfield
We will use DH subject_geographic for this field. The source of subject_geographic in DH is FAST-Geographic, so we can add that as the `identifier`.

sharpattack:
Use subject_geographic to populate locations metadata. closes #891

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -157,7 +157,7 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
       "formats": gf.mime_type,
       "version": version(gf.content),
       "rights": rights(gf.rights),
-      "locations": {"features": gf.based_near.present? ? gf.based_near.shift.split("', ").map{ |location| {place: location.gsub("'", "")} } : []},
+      "locations": {"features": gf.subject_geographic.present? ? gf.subject_geographic.map{ |location| {place: location} } : []},
       "funding": funding(gf.id)
     }
   end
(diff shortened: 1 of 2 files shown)


sharpattack:
Unfortunately, `subject_geographic` is stored with the text value only. So we are not able to get it's identifier to add.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -157,7 +157,7 @@ def invenio_metadata(gf)
       "formats": gf.mime_type,
       "version": version(gf.content),
       "rights": rights(gf.rights),
-      "locations": {"features": gf.based_near.present? ? gf.based_near.shift.split("', ").map{ |location| {place: location.gsub("'", "")} } : []},
+      "locations": {"features": gf.subject_geographic.present? ? gf.subject_geographic.map{ |location| {place: location.gsub("'", "")} } : []},
```
fenekku, line 160: Curious: Why do we have to do `location.gsub("'", "")`? (Other characters to remove?)

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -178,6 +178,7 @@
     subject { invenio_rdm_record_converter.to_json }
 
     it do
+      #binding.pry
```
fenekku, line 181: can remove

fenekku (approved):
Questions/minor things, once resolved we can merge.

app/models/invenio_rdm_record_converter.rb
sharpattack: Ah, good catch. We do not need it, I missed removing this.

spec/models/invenio_rdm_record_converter_spec.rb
sharpattack: Whoops, thank you. 🤦

Meowcenary (approved):

PR #898, opened 2021-07-28
[#895] Add Owner Email and Netid to Extras

### LINKED ISSUE #895: Export owner email + netid in extras
We need a way to tell who is the owner of the record and in InvenioRDM each user is uniquely identified by their email, but they might login with their netid in the future. So we need both at migration time.

Let's generate an extra field in the export that would hold that information:

```json
{
  "record": ...,
  "file": ...,
  "extras": {
    "owner": {
      "netid": ...,
      "email": ...
    },
  ...
  }
}
```

**Context**

The `"provenance"` field (which will not be supported in the LTS as far as I can tell) doesn't hold the information we (at Northwestern) need to associate a user to a record on migration. Usually a record's owner is determined by looking at the logged user making the deposit in InvenioRDM. But on migration we need to create that user first (no matter how we dice it).

I've also noticed that the `provenance` subfields for all records I've looked at are null:

```json
"provenance": {
  "created_by": {
    "user": null,
  },
  "on_behalf_of": {
    "user": null
  }
}
```
so we probably don't want to use that way to get the information.

sharpattack:
The depositor for a GenericFile is used as the 'owner'. The netid is
stored as 'username' for User. Look them up and extract the netid and
email and add to Extras field. closes #895

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -63,9 +63,28 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   end
 
   def extra_data(generic_file)
+    data = {}
+
     if !generic_file.based_near.empty?
+        data["presentation_location"] = generic_file.based_near
+    end
+    data["owner"] = owner_info(generic_file)
+
+    data
+  end
+
+  def owner_info(generic_file)
+    user = User.find_by(username: generic_file.depositor)
+
+    if user
       {
-        "presentation_location": generic_file.based_near
+        "netid": user.username,
+        "email": user.email
+      }
+    else
+      {
+        "netid": "unknown",
+        "email": "unknown"
       }
     end
   end
--- a/spec/models/invenio_rdm_record_converter_spec.rb
+++ b/spec/models/invenio_rdm_record_converter_spec.rb
@@ -162,7 +162,11 @@ RSpec.describe InvenioRdmRecordConverter do
         "content_path": generic_file_content_path
       },
       "extras": {
-        "presentation_location": generic_file.based_near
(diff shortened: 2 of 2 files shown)


Meowcenary (approved):
Looks good to me!

app/models/invenio_rdm_record_converter.rb
```diff
@@ -63,9 +63,28 @@ def generic_file_content_path(checksum)
...
+    if user
+      {
+        "netid": user.username,
+        "email": user.email
+      }
+    else
       {
-        "presentation_location": generic_file.based_near
+        "netid": "unknown",
+        "email": "unknown"
```
fenekku, line 87: :heavy_check_mark: I will make the Prsim admin user be the owner for those.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -275,15 +279,17 @@
 
   let(:expected_presentation_location) {
```
fenekku, line 280: Let's rename that to `:expected_extras`.

fenekku (approved):
Small rename for accuracy and then you can merge :+1:

app/models/invenio_rdm_record_converter.rb
sharpattack: Okay, sounds good to me. I am pretty sure this `"unknown"` case will be unlikely. Users can't change their usernames, or change the GenericFile's depositor. But, just in case.

spec/models/invenio_rdm_record_converter_spec.rb
sharpattack: Yes, I missed this one. Thank you for catching!

PR #899, opened 2021-07-30
896 record permissions

### LINKED ISSUE #896: Export record permissions
Although InvenioRDM doesn't support per user permissions yet (it will with communities), we will have to export the DigitalHub ones eventually.

The kind of information we want to capture is

- what permission(s)
- given to whom -- user(s) identified by netid(s) / email(s)

A reasonable structure for now is:

```json
{
  ...,
  "extras": {
    "permissions": {
      "edit": ["<user identifier>", ...]
      "view": [...]
    }
  }
} 
```

This is not a blocker for the initial export apriori. If we decide to wait to import when communities exist, then this should be part of that import. By then, the structure of these kinds of permissions will also be established and we will be able to use them. If we don't wait, the importer should be able to map that basic information from the `extras.permisions` into what InvenioRDM wants.

sharpattack:
Export the permissions for each GenericFile. Added to the "extras"
section, for each permission type(read/edit), store the user's
netid(username). closes #896

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -36,23 +36,24 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   # @param [GenericFile] generic_file file to be converted for export
   def initialize(generic_file=nil)
     return unless generic_file
+    @generic_file = generic_file
 
-    @record = record_for_export(generic_file)
-    @file = filename_and_content_path(generic_file)
-    @extras = extra_data(generic_file)
+    @record = record_for_export
+    @file = filename_and_content_path
+    @extras = extra_data
   end
 
   def to_json(options={})
-    options[:except] ||= ["memoized_mesh", "memoized_lcsh"]
+    options[:except] ||= ["memoized_mesh", "memoized_lcsh", "generic_file"]
     super
   end
 
   private
 
-  def filename_and_content_path(generic_file)
+  def filename_and_content_path
     {
-      "filename": generic_file.filename,
-      "content_path": generic_file_content_path(generic_file.content.checksum.value)
+      "filename": @generic_file.filename,
+      "content_path": generic_file_content_path(@generic_file.content.checksum.value)
     }
   end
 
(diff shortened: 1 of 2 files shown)


app/models/invenio_rdm_record_converter.rb
```diff
@@ -36,23 +36,24 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   # @param [GenericFile] generic_file file to be converted for export
   def initialize(generic_file=nil)
     return unless generic_file
+    @generic_file = generic_file
```
Meowcenary, line 39: Nice refactor here. It definitely makes sense as an instance variable.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -89,13 +91,23 @@ def owner_info(generic_file)
     end
   end
 
-  def record_for_export(generic_file)
+  def file_permissions
+    permission_data = Hash.new { |h,k| h[k] = [] }
```
Meowcenary, line 95: Edit: `Hash.new` seems right for here.

This just returns an empty hash in IRB/Rails console :

```
2.3.3 :001 > Hash.new{ |h, k| h[k] = [] }
 => {}
```

I say we replace it with either `Hash.new` or if you want to initialize a hash with an empty array in it you could do something like `hash = {key: []}`

app/models/invenio_rdm_record_converter.rb
```diff
@@ -89,13 +91,23 @@ def owner_info(generic_file)
...
+
+    @generic_file.permissions.each do |permission|
```
Meowcenary, line 97: Ah, you just construct it here. I'll update my above comment, but I think in this case you should just go with `Hash.new`

Meowcenary (approved):
Small review comment, but otherwise looks good to me! Thanks for handling this, I really did not know what I was doing and had gone down a totally incorrect path.

app/models/invenio_rdm_record_converter.rb
sharpattack: It might be my ruby version, but when I try just using `Hash.new`, I was not able construct the value to be an array on the fly. The `h[k] = []` part ensures that the value is an array, so I can use the shovel operator if the data didn't exist yet. Though, yes I could do `hash = {"read": [], "edit": []}`, I want to cover any types of permissions I might not know about.
``` ruby
[1] pry(main)> data = Hash.new
=> {}
[2] pry(main)> data["fizz"] << "buzz"
NoMethodError: undefined method `<<' for nil:NilClass
from (pry):2:in `<main>'
[3] pry(main)> data = Hash.new { |h,k| h[k] = [] }
=> {}
[4] pry(main)> data["fizz"] << "buzz"
=> ["buzz"]
```

Meowcenary: That's interesting. I wonder what's going on under the hood here? I checked myself and it's working as you described.

sharpattack:
> I really did not know what I was doing and had gone down a totally incorrect path.

No such thing! Nothing was broken, the code works great, we are making progress, and I am sure there are a million other things we can improve. I am just thankful for your review 🙌 

Going to deploy this shortly while trying to upgrade passenger.

PR #911, opened 2021-08-18
[#905] Export Uri As Subject Ids

### LINKED ISSUE #905: Export URI as subject ids
For MeSH the id is now of the form: `https://id.nlm.nih.gov/mesh/<D-id>`.

For LCSH they would be: `https://id.loc.gov/authorities/subjects/<sh-id>`

Meowcenary:
Update value that is exported for mesh and lcsh header ids. closes #905


spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -6,9 +6,9 @@
   let(:contributor_user) { FactoryGirl.create(:user, username: "contributor_user", formal_name: "User, Contributor",  display_name: 'Contributor User') }
   let(:assistant) { FactoryGirl.create(:user, username: "ast9876") }
   let(:mesh_term) { "Vocabulary, Controlled" }
-  let(:expected_mesh_pid) { "D018875" }
+  let(:expected_mesh_id) { ::HeaderLookup::MESH_ID_URI + "D018875" }
```
fenekku, line 9: Just curious: why this syntax `::<class>::<attribute>` and not `<class>.<attribute>` in ruby?

fenekku (approved):

sharpattack (approved):
The changes here are the right adjustments for this task/PR, so I will approve.

While I was taking a closer look at the tests, and the expected results, I found that things are not what they seem. We need to make changes to the LCSH and MESH lookup, to either match the result with what we query, or use the value that is returned. I will open a new ticket with more details.

spec/models/invenio_rdm_record_converter_spec.rb
Meowcenary: <class>.<attribute> is for calling class methods. ::<class>::<attribute> is specifically for getting a constant defined in a class.

PR #912, opened 2021-08-18
[#909] Export Human Readable Keywords In Subjects

### LINKED ISSUE #909: Export: human readable keywords in subjects
The "subjects" field has changed a little in the past release with respect to keywords (terms that are not part of an official vocabulary like MeSH). A human-readable label should be passed for the `subject` subfield.

They should be passed as follows e.g.,:

```json
{
  "subjects": [
    {
      "subject": "Image Analysis"
    }
  ]
}
```

https://inveniordm.docs.cern.ch/reference/metadata/#subjects-0-n 

(In the sample records we had for instance: `galter-keyword-image-analysis`)

Meowcenary:
Update subjects field to use human readable label for terms that are not part of an official vocabulary. closes #909


fenekku (approved):
LGTM

PR #917, opened 2021-08-20
[#908] Export Funding Should Be An Array

### LINKED ISSUE #908: Export: "funding" should be an array
In the latest sample export (2021-08-05), "funding" is given as an object, but should be an array.

https://inveniordm.docs.cern.ch/reference/metadata/#funding-references-0-n

Meowcenary:
closes #908


app/models/invenio_rdm_record_converter.rb
```diff
@@ -441,6 +441,6 @@ def normalize_date(date_string)
   end
 
   def funding(file_id)
-    @@funding_data[file_id] || {}
+    @@funding_data[file_id] || [{}]
```
fenekku, line 444: minor: could just be empty array (no need for hash in it)

fenekku (approved):
Thanks!

PR #928, opened 2021-08-31
[#913] Update Perform And Parse

### LINKED ISSUE #913: Update perform_and_parse
For both LCSH and MeSH query lookups. Right now, we are setting the id/uri to the first returned result. It may not be the query. 

i.e. Using "cancer" vocabulary, the current `#perform_and_parse_mesh_query` will return "Abdomen--Cancer" as the first result, and we end up pairing "cancer" with the id of "Abdomen--Cancer".

Meowcenary:
Update LCSH and MeSH header lookup to use local file of terms before
checking online. Terms are stored in subject_lcsh.yml and
subject_mesh.yml. The subject_lcsh.yml file is created the from
parse_lcsh_file.rb script. closes #913


app/models/header_lookup.rb
```diff
@@ -10,24 +10,57 @@ class HeaderLookup
...
+    @@memoized_mesh ||= read_memoized_headers(MEMOIZED_MESH_FILE)
+    @@memoized_lcsh ||= read_memoized_headers(MEMOIZED_LCSH_FILE)
   end
 
   def pid_lookup_by_scheme(term="", scheme="")
     if term.blank? || scheme.blank?
       nil
     elsif scheme == :mesh
-      @@memoized_mesh[term] || mesh_term_pid_lookup(term)
+      @@memoized_mesh[term] || mesh_term_pid_local_lookup(term) || mesh_term_pid_lookup(term)
```
Meowcenary, line 30: start with memoized, then do a local lookup, and finally do a search

app/models/header_lookup.rb
```diff
@@ -10,24 +10,57 @@ class HeaderLookup
...
+    @@searchable_lcsh_terms.each do |term_json|
+      if term_json["subject"].downcase == lcsh_term.downcase
+        lcsh_id = term_json["id"]
+        @@memoized_lcsh[lcsh_term] = lcsh_id
+        File.write(MEMOIZED_LCSH_FILE, @@memoized_lcsh)
+        return lcsh_id
+      end
+    end
+
+    nil
```
Meowcenary, line 61: if nothing is found, return nil so that the `||` statement will continue on to use a web api lookup

memoized_lcsh.txt
```diff
@@ -1 +1 @@
-{"Semantic Web"=>"https://id.loc.gov/authorities/subjects/sh2002000569", "cancer"=>"https://id.loc.gov/authorities/subjects/sh85000095"}
\ No newline at end of file
+{"Semantic Web"=>"https://id.loc.gov/authorities/subjects/sh2002000569", "cancer"=>"https://id.loc.gov/authorities/subjects/sh85000095", "Cancer"=>"http://id.loc.gov/authorities/subjects/sh85019492"}
```
Meowcenary, line 1: These caches had to be destroyed to make sure we were getting correct values.

parse_lcsh_file.rb
```diff
@@ -0,0 +1,37 @@
...
+lcsh_file.each_line do |line|
+  line_graph_json = JSON(line)["@graph"]
+
+  line_graph_json.each do |node|
+    if node["@type"] == "skos:Concept" && node["skos:prefLabel"].present?
+      id = node["@id"]
+
+      if !memoized_ids.include?(id)
+        memoized_ids << id
+        lcsh_string = %(- id: #{id}\n  scheme: lcsh\n  subject: "#{find_subject(node["skos:prefLabel"])}"\n)
```
fenekku, line 32: InvenioRDM double-dips on the scheme and uses it as a human readable label too so we want it cased appropriately.

```suggestion
        lcsh_string = %(- id: #{id}\n  scheme: LCSH\n  subject: "#{find_subject(node["skos:prefLabel"])}"\n)
```

On another note, I guess the stdlib YAML would have wanted the whole document in memory before dumping it (https://ruby-doc.org/stdlib-2.5.1/libdoc/yaml/rdoc/YAML.html). This may or may not have been ideal given the size of data we are dealing with. Alternative would have been to use the YAML stream writing and remove the "---" after the fact. Messy either way. Even in Python I haven't found a decent YAML array streaming library...  This works, so :+1:

fenekku (commented):
Thanks for that. Just small things.

memoized_lcsh.txt
Meowcenary: Good point. I did a test run of the exporter to make sure things were not breaking yesterday. I'll plan on doing that again after finishing all my changes then I can copy the cache file to my local instance and include it in the commit.

carsonicator (approved):
Thanks @Meowcenary

fenekku (approved):

PR #931, opened 2021-09-01
[#924] Add ID to Dates

### LINKED ISSUE #924: [Export] Make dates have id
The `dates` field format should be as documented here https://inveniordm.docs.cern.ch/reference/metadata/#dates-0-n (no need to pass the `title` subfield though*). So something like:

```json
{
  "dates": [{
    "date": "1939/1945",
    "type": {
      "id": "other",
    },
    "description": "A date"  // optional
  }],
}
```

The `id` must be taken from: https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/fixtures/data/vocabularies/date_types.yaml

`*`: This is admittedly not good documentation. The Metadata Reference section is really meant for librarians to understand what is stored in their instance. The REST API has sligthly different input / output details that are not captured there.

sharpattack:
For 'type'. closes #924

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -180,7 +180,7 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
       "publication_date": format_publication_date(@generic_file.date_created.shift || @generic_file.date_uploaded.to_s),
       "subjects": SUBJECT_SCHEMES.map{ |subject_type| subjects_for_scheme(@generic_file.send(subject_type), subject_type) }.compact.flatten,
       "contributors": contributors(@generic_file.contributor),
-      "dates": @generic_file.date_created.map{ |date| {"date": date, "type": {"id": "other"}, "description": "When the item was originally created."} },
+      "dates": @generic_file.date_created.map{ |date| {"date": normalize_date(date), "type": {"id": "created"}, "description": "When the item was originally created."} },
       "languages": @generic_file.language.map{ |lang| lang.present? && lang.downcase == ENGLISH ? {"id": "eng"} : nil }.compact,
       "identifiers": ark_identifiers(@generic_file.ark),
       "related_identifiers": related_identifiers(@generic_file.related_url),
@@ -375,6 +375,11 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
 
   def normalize_date(date_string)
     split_date = date_string.split(/[-,\/ ]/).map(&:downcase)
+    # date format starts with month first
+    if (!split_date.blank? && split_date[0].length < 3)
+      split_date = rearrange_year(split_date)
+    end
+
     month_names = (split_date & MONTHNAMES)
     abbr_month_names = (split_date & ABBR_MONTHNAMES)
 
@@ -438,6 +443,19 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
     else
       ""
     end
+  end
+
+  def rearrange_year(date_array)
+    if date_array[2].length == 2
+      # drawing the line at 22 as the year for when we no longer refer to 1900s
+      if date_array[2].to_i > 22
+        date_array[2] = "19#{date_array[2]}"
+      else
+        date_array[2] = "20#{date_array[2]}"
+      end
+    end
+
+    [date_array[2], date_array[0], date_array[1]]
   end
 
   def funding(file_id)
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -446,16 +446,19 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   end
 
   def rearrange_year(date_array)
-    if date_array[2].length == 2
-      # drawing the line at 22 as the year for when we no longer refer to 1900s
-      if date_array[2].to_i > 22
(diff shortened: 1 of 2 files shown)


app/models/invenio_rdm_record_converter.rb
```diff
@@ -180,7 +180,7 @@ def invenio_metadata
       "publication_date": format_publication_date(@generic_file.date_created.shift || @generic_file.date_uploaded.to_s),
       "subjects": SUBJECT_SCHEMES.map{ |subject_type| subjects_for_scheme(@generic_file.send(subject_type), subject_type) }.compact.flatten,
       "contributors": contributors(@generic_file.contributor),
-      "dates": @generic_file.date_created.map{ |date| {"date": date, "type": "other", "description": "When the item was originally created."} },
+      "dates": @generic_file.date_created.map{ |date| {"date": date, "type": {"id": "other"}, "description": "When the item was originally created."} },
```
fenekku, line 183: Oh this is always the date of creation? Then, we should use this id: https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/fixtures/data/vocabularies/date_types.yaml#L21

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -104,7 +104,7 @@
             },
             "role": {"id": InvenioRdmRecordConverter::ROLE_OTHER}
           }],
-          "dates": [{"date": "2021-1-1", "type": "other", "description": "When the item was originally created."}],
+          "dates": [{"date": "2021-1-1", "type": {"id":"other"}, "description": "When the item was originally created."}],
```
fenekku, line 107: The date format should follow EDTF level 0 https://inveniordm.docs.cern.ch/reference/metadata/#dates-0-n which means double digit resolution i.e. it should be `"2021-01-01"`.

fenekku (commented):
More tweaks :paintbrush:

Meowcenary:
Seems like @fenekku has covered all the necessary changes. Looks pretty straightforward to me so approving pending changes.

Meowcenary (approved):

spec/models/invenio_rdm_record_converter_spec.rb
sharpattack: Will the import still allow any dates? Or is it strictly EDTF level 0? This a free text field in DigitalHub, unfortunately.

fenekku: No, the import only allows EDTF level 0 dates.

sharpattack:
@saragon02 There are dates formatted as "mm/dd/yy". Because we need the format to be in EDTF level 0, the year has to be "YYYY", fully 'spelled' out. Is it okay if we draw the line at... 22? So "yy" in the range of 00-22 will be in the 2000s, and the range of 23-99 will be in the 1900s? A closer look at the recent metadata dump sent to Liping may help answer this.

saragon02:
@sharpattack If the purpose is for automating conversion of Date Created dates to YYYY format, I think that may not work because I'm pretty confident we'll have some pre-1900 creation dates for some items (Civil War era things, certain photos & letters, etc.)

sharpattack:
Thanks to @saragon02 for the help! +1 to using OpenRefine. I was using excel to examine a recent metadata dump, and excel decides to reformat dates to their liking. There are no dates in formatted with two digits, as "yy". 
@fenekku and @Meowcenary the changes are now ready for review.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -440,6 +445,19 @@ def normalize_date(date_string)
     end
   end
 
+  def rearrange_year(date_array)
+    if date_array[2].length == 2
+      # drawing the line at 22 as the year for when we no longer refer to 1900s
+      if date_array[2].to_i > 22
```
Meowcenary: I want to put the 22 into a constant or something similar to describe what it's doing, but I honestly, I can't really think of a good name. The comment is very helpful though.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -375,6 +375,11 @@ def format_publication_date(publication_date)
 
   def normalize_date(date_string)
     split_date = date_string.split(/[-,\/ ]/).map(&:downcase)
+    # date format starts with month first
+    if (!split_date.blank? && split_date[0].length < 3)
```
Meowcenary, line 379: Thanks for figuring this out! I was having a really hard time with these kinds of dates previously.

Meowcenary (approved):
Looks good to me Austin. Thanks for working on this.

fenekku (approved):
So the case of a year with 2 digits never really happened?

app/models/invenio_rdm_record_converter.rb
sharpattack: Haha, well comment or constant, we don't need this anymore. There are no dates with just "yy". They all include the century, "YYYY".

app/models/invenio_rdm_record_converter.rb
sharpattack: Thanks to you for getting this started! I only added a small piece, while you did the heavy lifting with the dates.

sharpattack:
*sigh* There were no years with 2 digits. Thankfully, though.

fenekku:
ok

PR #935, opened 2021-09-08
[#925] Export Funding Fixes

### LINKED ISSUE #925: [Export] Funding fixes
`funding` has to adhere to https://inveniordm.docs.cern.ch/reference/metadata/#funding-references-0-n:

- if no funding, then should be `"funding": []` (or no "funding" key at all) as opposed to `"funding": [{}]` (per _Originally posted by @fenekku in https://github.com/galterlibrary/digital-repository/pull/917#discussion_r693169816_ )
- `funder.name` is required (cannot be empty if a `funder` object is provided)
- `award.title` is required (cannot be empty if an `award` object is provided)
- `award.number` is required (cannot be empty if an `award` object is provided)

The following is incorrect for example:

```json
        {
          "funder": {
            "name": "National Library of Medicine (NLM)",
            "identifier": "0060t0j89",
            "scheme": "ror"
          },
          "award": {
            "title": "",  // should not be empty
            "number": "F37 LM009568 ",
            "identifier": "",
            "scheme": ""
          }
        }
```

We may have to copy award number into award title if there is no appropriate value at export time (and other similar patches).

Meowcenary:
WIP - issues with header lookup are breaking specs, but should probably be addressed separately of changes done here. Specs should, IMO, be modified slightly to use more specific terms for now (e.g "abdominal cancer" instead of "cancer"). Will discuss at standup tomorrow.

Update funding to fit InvenioRDM standard for September 2021.
closes #925


app/models/invenio_rdm_record_converter.rb
```diff
@@ -441,6 +442,21 @@ def normalize_date(date_string)
...
   def funding(file_id)
-    @@funding_data[file_id] || [{}]
+    funding_sources = @@funding_data[file_id]
+
+    if funding_sources.blank?
+      return []
+    end
+
+    funding_sources.map do |source|
+      source == BLANK_FUNDER_SOURCE
```
fenekku, line 452:
```suggestion
      source = BLANK_FUNDER_SOURCE
```

app/models/invenio_rdm_record_converter.rb
```diff
@@ -441,6 +442,21 @@ def normalize_date(date_string)
...
+
+      # if the title is blank, replace it with the award number
+      if source[:award][:title].blank?
+        source[:award][:title] = source[:award][:number]
+      end
```
fenekku, line 457: I guess we'll have to see if there are funding entries that are missing other required fields when we import. (`funder.name`, `award.number`). If we do we will have to revisit.

fenekku (commented):
Let's discuss the lookup tomorrow: I think I am still wondering why we would be doing a remote query.

app/models/invenio_rdm_record_converter.rb
Meowcenary: Oh jeez, there was supposed to be logic here. I think my brain puttered out from being tired around the time I was pushing this up.

app/models/invenio_rdm_record_converter.rb
Meowcenary: I was looking at funding data from production and I think everything else is there, but we can run a script to make sure.

app/models/invenio_rdm_record_converter.rb
Meowcenary: Okay, this has been updated to the proper `return [] if source == BLANK_FUNDER_SOURCE`

Meowcenary:
PR has been updated and is ready for review and merging. Tagged @sharpattack since apparently I forgot the first time around :(

app/models/invenio_rdm_record_converter.rb
Meowcenary: Wait, this is incorrect. Return will actually exit the function. Ugh, need to update again.

Meowcenary:
Oof, this was really terrible due to a lot of memoized values that had changed and needed to be tracked down. I think we will need to change ALL the memoized mesh headers from `http` to `https`, but I'm pretty sure the lcsh headers are correctly all stored with `https`.

memoized_lcsh.txt
```diff
@@ -1 +1 @@
-{"Sailboats"=>"http://id.loc.gov/authorities/subjects/sh85116520", "Periodical cicada"=>"http://id.loc.gov/authorities/subjects/sh85099887", "Saguenay River (Quebec)"=>"https://id.loc.gov/authorities/subjects/sh86006143", "Social stratification"=>"http://id.loc.gov/authorities/subjects/sh99000315", "Originality"=>"http://id.loc.gov/authorities/subjects/sh85095646", "Sailing"=>"http://id.loc.gov/authorities/subjects/sh85116532", "Marsden"=>"https://id.loc.gov/authorities/subjects/sh99014562", " Sam (Fictitious character)"=>"https://id.loc.gov/authorities/subjects/sh2008000533", "Molds (Fungi)"=>"http://id.loc.gov/authorities/subjects/sh85086566", "Lindens"=>"http://id.loc.gov/authorities/subjects/sh85077148", "Statues"=>"http://id.loc.gov/authorities/subjects/sh2001008682", "Test family"=>"http://id.loc.gov/authorities/subjects/sh92004443", "Semantic Web"=>"https://id.loc.gov/authorities/subjects/sh2002000569", "Cancer"=>"http://id.loc.gov/authorities/subjects/sh85019492"}
\ No newline at end of file
+{"Semantic Web"=>"http://id.loc.gov/authorities/subjects/sh2002000569", "Abdomen--Cancer"=>"http://id.loc.gov/authorities/subjects/sh85000095"}
```
fenekku, line 1: I guess LCSH used the URI as an identifier and is stuck with the http scheme until they update it and add another URI.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -184,6 +184,7 @@
     subject { invenio_rdm_record_converter.to_json }
 
     it do
+      puts subject, json
```
fenekku, line 187: minor: can be removed right?

fenekku (approved):
An ultra minor stray `puts` is the only thing left. 

Put this one in the ground after.

PR #940, opened 2021-09-14
[#939] Update Header Lookup To Only Check Local Sources

### LINKED ISSUE #939: Update Header Lookup To Only Check Local Sources
Any headers that we are looking up should be in the local mesh yml file. If the header can not be found there instead of looking it up through the API we will have it output a message saying what the term was and what scheme it used (in case we decide to use this process for LCSH headers as well).

Meowcenary:
Remove calls to mesh and lcsh lookup apis. When a subject lookup can not
find a term locally, output information for looking up manually.
closes #939

### DIFF
--- a/app/models/header_lookup.rb
+++ b/app/models/header_lookup.rb
@@ -27,9 +27,9 @@ class HeaderLookup
     if term.blank? || scheme.blank?
       nil
     elsif scheme == :mesh
-      @@memoized_mesh[term] || mesh_term_pid_local_lookup(term) || mesh_term_pid_lookup(term)
+      @@memoized_mesh[term] || mesh_term_pid_local_lookup(term) || nil
     elsif scheme == :lcsh
-      @@memoized_lcsh[term] || lcsh_term_pid_local_lookup(term) || lcsh_term_pid_lookup(term)
+      @@memoized_lcsh[term] || lcsh_term_pid_local_lookup(term) || nil
     else
       nil
     end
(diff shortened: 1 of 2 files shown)


app/models/header_lookup.rb
```diff
@@ -27,9 +27,9 @@ def pid_lookup_by_scheme(term="", scheme="")
     if term.blank? || scheme.blank?
       nil
     elsif scheme == :mesh
-      @@memoized_mesh[term] || mesh_term_pid_local_lookup(term) || mesh_term_pid_lookup(term)
```
Meowcenary, line 30: it seemed that the easiest thing to do here was to simply return nil and then react to that in the function that calls this one. I left the original functions `mesh_term_pid_lookup` and `lcsh_term_pid_lookup` in case we need to use those again, but we can remove them if need be since we can always look at old versions on git.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -293,6 +293,8 @@ def subjects_for_scheme(terms, scheme)
         {id: pid}
       elsif scheme == :subject_name || scheme == :tag
         {subject: term}
+      else
+        puts "------\nUnable to map subject\nFile Id: #{@generic_file.id} Term: #{term} Scheme: #{scheme}\n------"
```
Meowcenary, line 297: This seemed like the easiest way to do this without adding yet another file write out to this, but if we would prefer to write out to a file that would be an easy change to make.

fenekku: we can always redirect the stdout to file anyway, so that's ok.

fenekku (approved):
Should be faster and more accurate now :1st_place_medal:

app/models/invenio_rdm_record_converter.rb
Meowcenary: Oh, good point.

PR #947, opened 2021-09-24
933 export original_checksum

### LINKED ISSUE #933: [Export] Add file MD5 in export metadata
In the exported json metadata for a record, we should pass the record's file MD5 hash as such:

```json
{
  ...
  "file": {
    "filename": [
      "IMGA_1.0.0_distribute.zip"
    ],
    "content_path": "/hydra-jetty/fcrepo4-data/fcrepo.binary.directory/2c/ce/b8/2cceb8d6f0fc2ad14d38bf94d298dd5c4dfdadca",
    "md5": "398b1616d52178e9b38a00d88774c579"
  }
```

### LINKED ISSUE #936: Export File Content Path Fix
File's content paths are created based on their SHA1 checksum. However, if a duplicate file has been uploaded, the checksum is the same, but the path already exists. In this case, the path then is based on the GenericFile's id. We need to check the paths, and use the correct one for export.

sharpattack:
- Add checksum path tester script
- Add original_checksum to file export
- closes #933 
- closes #936


app/models/invenio_rdm_record_converter.rb
```diff
@@ -51,10 +52,11 @@ def to_json(options={})
 
   private
 
-  def filename_and_content_path
+  def file_info
     {
       "filename": @generic_file.filename,
-      "content_path": generic_file_content_path(@generic_file.content.checksum.value)
+      "content_path": generic_file_content_path(@generic_file.content.checksum.value),
+      "original_checksum": @generic_file.original_checksum
```
sharpattack, line 59: From what I checked, `original_checksum` matched the file's md5 checksum. 🤷‍♂️

fenekku: wow ok, they generate a SHA1 and an MD5 from every file then...

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -157,7 +157,8 @@
       },
       "file": {
         "filename": generic_file.filename,
-        "content_path": generic_file_content_path
+        "content_path": generic_file_content_path,
+        "original_checksum": generic_file.original_checksum
```
fenekku, line 161: Could we have a test for the `original_checksum` actually? To make sure that the original_checksum is produced correctly.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -51,10 +52,11 @@ def to_json(options={})
 
   private
 
-  def filename_and_content_path
+  def file_info
```
Meowcenary, line 55: Good name change. Less verbose and gets the point across fine.

Meowcenary (approved):
Looks good to me

spec/models/invenio_rdm_record_converter_spec.rb
sharpattack: We actually did add this when we first added `#generic_file_content_path` as seen [here](https://github.com/galterlibrary/digital-repository/blob/a5feaf4b39835e1b3709b03a004380b2c30e9d7d/spec/models/invenio_rdm_record_converter_spec.rb#L277), but let me know if we want more edge cases than just one.

app/models/invenio_rdm_record_converter.rb
sharpattack: @fenekku seems so, I wonder what else there could be...

spec/models/invenio_rdm_record_converter_spec.rb
fenekku: I guess I thought the spec tests used an actual file: then we could have tested our expectation that `original_checksum` is the MD5 checksum of the actual file. It's fine if we don't since the test might not have been representative of production anyway and it's better to get feedback from production first.

Thanks!

fenekku (approved):
Thanks. Ready to be merged!

PR #964, opened 2021-11-12
[#963] Export Collection Info

### LINKED ISSUE #963: Export Collection Info
Export new key "communities" with collection info a record belongs to in DigitalHub. 

```
{
  "record": {
    ...
  },
  "file": {
    ...
  },
  "extras": {
    ...
  },
  "communities": [
    {
      "id": <id>,
      "url": <url>,
      "title": <title>
    }
  ]
}
```

sharpattack:
Export collection info for records. Includes all nested/parent
collection paths. closes #963

Also removes unused libraries in InvenioRdmRecordConverter.


fenekku (approved):
I highly suspect this will be hard/slow on the DB/machine, but try it out if you didn't. Then you can use the speedup tactic outlined.

lib/scripts/collection_list.rb
```diff
@@ -0,0 +1,32 @@
...
+end
+
+def recursive_collection_path(collection, path, collection_paths)
+  if collection.collections.empty?
+    collection_paths << path
+  else
+    collection.collections.each do |collection|
+      recursive_collection_path(
+        collection,
+        [{"title": collection.title, "id": collection.id}] + path,
```
Meowcenary, line 27: `{"title": collection.title, "id": collection.id}` shows up in both `#recursive_collection_path` and `#collections_path`. Also, `{"title": parent_collection.title, "id": parent_collection.id}` is basically the same idea, but for a parent collection. It should be wrapped into a function that describes what this array is:

```
def name_describing_data(collection)
  [{"title": collection.title, "id": collection.id}]
end
```

lib/scripts/collection_list.rb
```diff
@@ -0,0 +1,32 @@
...
+        collection_paths
+      )
+    end
+  end
+end
```
Meowcenary, line 32: I had some small changes that I think make this a little cleaner:

```
def collections_path(collection)
  if collection.collections.empty?
    collection_paths = [[name_describing_data(collection)]]
  else
    collection_paths = []
    path = [{name_describing_data(parent_collection)},
            {name_describing_data(collection)}]

    collection.collections.each do |parent_collection|
      recursive_collection_path(parent_collection, path, collection_paths)
    end

    collection_paths
  end

def recursive_collection_path(collection, path, collection_paths)
  if collection.collections.empty?
    collection_paths << path
  else
    collection.collections.each do |collection|
      path = [name_describing_data(collection)] + path
      recursive_collection_path(collection, path, collection_paths)
    end
  end
end

# I wasn't sure what this was exactly, so I just put a placeholder name
def name_describing_data(collection)
  {"title": collection.title, "id": collection.id}
end
```

lib/scripts/repo_export_in_batches.rb
```diff
@@ -20,14 +22,22 @@
   raise(RegistryError, "Model (#{converter[:model_class].name}) for conversion must be an ActiveFedora::Base") unless converter[:model_class].ancestors.include?(ActiveFedora::Base)
 end
 
+puts "---------\nCreating Collection Store\n---------"
+collection_store = {}
+Collection.find_each do |collection|
```
Meowcenary, line 27: Are we going to build the `collection_store` for each record every time we run the export or does it make sense to memoize this information somehow? Since this is something that is tied to the `Collection` model maybe we can add a field to the `Collection` table in the database with a callback on record update to recalculate the collections_path. If that's too much we could also just use a memoized file like we've done with other data. Either way, I think that's a lot faster than recursively checking every collection for collections.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -302,6 +307,135 @@
...
+        ]
+      }
+
+      let(:collection_store) {
+        {"community-1": [
+          [{"title": "Mom", "id": "parent-1"},
+           {"title": "Community", "id": "community-1"}],
+          [{"title": "Dad", "id": "parent-2"},
+           {"title": "Community", "id": "community-1"}]
+        ]}.with_indifferent_access
```
Meowcenary, line 425: Glad you were able to figure that out. I'm starting to wonder whether it makes more sense to just use hash rockets where you need to use characters incompatible with symbols in a hash or if there is some other solution. Something to look into some other time I guess.

sharpattack: Yea, but as we both agreed on, the test should not have done any transforming. It should have tested this as it is, and everything would have been alright. It's not our fault, but geeze it was a headache to realize what was happening.

sharpattack (commented):
I will add @Meowcenary suggested changes to the `collecrtion_list`, then notify reviewers the changes are up for another review.

lib/scripts/repo_export_in_batches.rb
Meowcenary: Oh okay. Since it is run only once let's just leave it as is.

lib/scripts/collection_list.rb
Meowcenary: I meant less describing the data literally and more describing what data represents. It looks like it's being used in for collection paths, but I don't know what exactly it is in a collection path. Is it the current collections path or path fragment?

carsonicator (approved):
Thanks @sharpattack

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -302,6 +307,136 @@
...
+
+      let(:collection_store) {
+        {"community-1": [
+           [{"title": "Community", "id": "community-1"}]
+         ],
+         "collection-1": [
+           [{"title": "Collection", "id": "collection-1"}]
+         ]
+        }.with_indifferent_access
+      }
```
sharpattack, line 357: I originally messed this `collection_store` up earlier.

sharpattack (commented):
@Meowcenary and @fenekku please review the latest. It should be good to go now.

lib/scripts/collection_list.rb
```diff
@@ -0,0 +1,37 @@
+def collections_path(collection)
+  starting_collection_data = collection_title_id_data(collection)
```
Meowcenary, line 2: Perfect. It's much more clear what this is being used for now.

lib/scripts/collection_list.rb
```diff
@@ -0,0 +1,37 @@
...
+  if collection.collections.empty?
+    collection_paths = [[starting_collection_data]]
+  else
+    collection_paths = []
+  end
```
Meowcenary, line 7: This isn't super important, but since there isn't a case where `collection.collections.empty? == true` will lead to the `collection.collections.each ...` doing anything wouldn't it make sense to move that loop into the else block like this:

```
else
  collection.collections.each do |parent_collection|
    recursive_collection_path(
      parent_collection,
      [collection_title_id_data(parent_collection),
       starting_collection_data],
      collection_paths
    )
  end

  collection_paths
end
```

It makes it immediately clear that if `collection.collections.empty? == true` nothing else is done.

Meowcenary (approved):
One small comment, but nothing that I think should prevent this from being merged.

sharpattack:
@fenekku and @Meowcenary please review the latest commit for changes. I will be fixing the test and rebasing soon, then request a final review when tests are green.

lib/scripts/collection_list.rb
```diff
@@ -1,37 +1,41 @@
-def collections_path(collection)
-  starting_collection_data = collection_title_id_data(collection)
-  if collection.collections.empty?
-    collection_paths = [[starting_collection_data]]
-  else
-    collection_paths = []
+def build_collection_store_data
```
sharpattack, line 1: For easier review of these changes, I will be renaming this file to correspond to this method in an upcoming commit.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -109,9 +109,7 @@ def list_collections
     collection_paths = []
 
     @generic_file.collections.each do |collection|
-      @collection_store[collection.id.to_sym].each do |path_item|
-        collection_paths << path_item
-      end
+      collection_paths << @collection_store[collection.id][:path]
```
sharpattack, line 112: I should change this to `@generic_file.collection_ids.each`

fenekku (commented):
Are the tests ready to be looked at?

lib/scripts/collection_list.rb
```diff
@@ -0,0 +1,41 @@
...
+        [collection_title_id_data(parent_collection_data)] + path,
+        starting_collection
+      )
+    end
+  end
+end
+
+def collection_title_id_data(collection)
+  {"title": collection[:title], "id": collection[:id]}
+end
```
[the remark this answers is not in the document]
sharpattack: Yes, this is the review comment I need. 🙌

sharpattack:
@fenekku class `CollectionStore` added, and tests updated.

app/models/collection_store.rb
```diff
@@ -0,0 +1,50 @@
...
+    @data = Hash.new
+  end
+
+  def build_collection_store_data
+    Collection.find_each do |c|
+      @data[c.id] = {
+        "id": c.id,
+        "title": c.title,
+        "collections": c.collection_ids,
+        "path": []
```
fenekku, line 14: minor: 
```suggestion
        "paths": []
```

fenekku (approved):
Awesome Austin! I know that was a long task and I am glad you got it through. We all learned  along the way :tada: .

This is good to be merged with or without that minor naming suggestion :+1: 

---

If this wet your appetite for algorithms and improvements, you may want to think about the following in the future or on your own: there is a lot of in-memory re-treading. If we have
```
Collection A ◄─┐                         │Collection D
               │                         │
               ├──Collection C ◄─────────┤
               │                         │
Collection B ◄─┘                         │Collection E  

```
calculating the paths from Collection E is pretty much the same thing as calculating the paths from Collection D (only the last segment differs). The recursive calls are repeated which can be a lot of work. It would be nice if we didn't have to re-compute all the time. In fact if we knew all the paths from Collection C already, computing Collection D would just be adding D to all of those (same thing respectively for E). 

So the trick is to 1- make sure Collection C is computed first and 2- the paths are stored for re-use. 

2- is easy: we already are saving paths per collection! We just need to use them.
1- is a matter of ordering: if we process all collections that have no parent first, then all collections that have those as parents and so on, we never have to recurse. The parent will always have all the paths from it.    

I am pretty sure this is what is called dynamic programming in algorithms classes. Anyway thanks again for all the work!

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -305,6 +307,119 @@
     end
   end
 
+  describe "#list_collections" do
```
Meowcenary, line 310: Something to start considering is how to avoid hitting the database during testing since it drastically slows down the time it takes to run the test suite. For this particular PR it's not a big deal and especially since we are in the process of sunsetting DH, but for future features and refactors on the Galter Website it's something we should strive for. I'll try to put together a little tech talk on it for the future, maybe something for the book club.

Meowcenary (approved):
Thanks for all the hard work on this!

sharpattack:
Thank you for the patience, @Meowcenary @carsonicator @fenekku

PR #967, opened 2021-12-06
[#125] Dont Export Records With Missing Dois

### LINKED ISSUE #965: [Export] Don't export records of the old GV Black collection
The "old GV Black collection" has to be identified. It contains some "combined" records that are not needed anymore e.g. generic_file_fq977t82f.json from the export. 

:warning: **WARNING** "combined" has become an overloaded terms in our discussions:
- There are records that have been created with the purpose of combining multiple files together and have a DOI. Those we want to keep.
- There are records with "combined" in the title. Those we don't want to keep typically.

Double-check and investigate this.

Meowcenary:
- Add #unexportable? function to GenericFile. 
- closes galterlibrary/digitalhub-2#125
- closes #965

### DIFF
--- a/app/models/generic_file.rb
+++ b/app/models/generic_file.rb
@@ -7,6 +7,11 @@ class GenericFile < ActiveFedora::Base
   include Galtersufia::GenericFile::FullTextIndexing
   include Galtersufia::GenericFile::MimeTypes
 
+  PUBLIC_PERMISSION = "http://projecthydra.org/ns/auth/group#public"
+  GV_BLACK_PHOTOGRAPH_SUB_COLLECTION_ID = "x346d4254"
+  GV_BLACK_COLLECTION_ID = "a4de96c9-7c6d-40d6-ad9e-cac8a24faad5"
+
+
   belongs_to :parent,
     predicate: ActiveFedora::RDF::Fcrepo::RelsExt.isPartOf,
     class_name: "Collection"
@@ -136,6 +141,24 @@ class GenericFile < ActiveFedora::Base
     citation = super
     add_doi_to_citation(citation)
   end
+
+  def unexportable?
+    collection_ids = collections.map(&:id)
+
+    # if record is publc, has a doi, and is either in the gv black photograph sub collection or is not in the greater gv black collection
+    public? && !doi.empty? && (collection_ids.include?(GV_BLACK_PHOTOGRAPH_SUB_COLLECTION_ID) || !collection_ids.include?(GV_BLACK_COLLECTION_ID))
+  end
+
+  def public?
+    permissions.each do |permission|
+      if permission.agent == PUBLIC_PERMISSION
+        return true
+      end
+    end
+
+    false
+  end
+  private :public?
 
   def json_presentation_terms
     GalterGenericFilePresenter.terms -
--- a/app/models/generic_file.rb
+++ b/app/models/generic_file.rb
@@ -143,10 +143,19 @@ class GenericFile < ActiveFedora::Base
   end
 
   def unexportable?
-    collection_ids = collections.map(&:id)
+    (open_access? && doi.empty?) || in_gv_black_collection?
+  end
 
-    # if record is publc, has a doi, and is either in the gv black photograph sub collection or is not in the greater gv black collection
-    public? && !doi.empty? && (collection_ids.include?(GV_BLACK_PHOTOGRAPH_SUB_COLLECTION_ID) || !collection_ids.include?(GV_BLACK_COLLECTION_ID))
+  def open_access?
+    self.visibility == InvenioRdmRecordConverter::OPEN_ACCESS
+  end
+
+  def in_gv_black_collection?
+    parent_collections_ids = self.collection_ids
+
+    # if broader restrictions -> add additional GV black collections
+    # if less restrictive -> use only GV black photograph sub collection
+    parent_collections_ids.include?(GV_BLACK_PHOTOGRAPH_SUB_COLLECTION_ID, GV_BLACK_COLLECTION_ID)
   end
 
   def public?
--- a/spec/models/generic_file_spec.rb
+++ b/spec/models/generic_file_spec.rb
@@ -625,4 +625,156 @@ RSpec.describe GenericFile do
       end
     end
   end
+
+  let(:public_permission_double) { double("Hydra::AccessControls::Permission") }
+  let(:gv_black_photo_sub_collection_double) { double("Collection") }
+  let(:gv_black_collection_double) { double("Collection") }
+  let(:random_doi) { "10.18131/g3-nqgv-4863" }
+  let(:blank_doi) { "" }
+  let(:public_record) { GenericFile.new(doi: []) }
+  let(:private_record) { GenericFile.new(doi: []) }
+
+  # if record is publc, has a doi, and is either in the gv black photograph sub collection or is not in the greater gv black collection
+  describe "#unexportable?" do
+    before do
+      allow(public_permission_double).to receive(:agent).and_return(GenericFile::PUBLIC_PERMISSION)
+
+      allow(gv_black_photo_sub_collection_double).to receive(:map).and_return([GenericFile::GV_BLACK_PHOTOGRAPH_SUB_COLLECTION_ID])
+      allow(gv_black_photo_sub_collection_double).to receive(:id)
+
+      allow(gv_black_collection_double).to receive(:map).and_return([GenericFile::GV_BLACK_COLLECTION_ID])
+      allow(gv_black_collection_double).to receive(:id)
+    end
+
+    context "record is public" do
+      before do
+        # GenericFile validates the type of permissions so simply passing a double to the constructor does not work
+        public_record.stub(:permissions).and_return([public_permission_double])
+      end
+
+      context "has doi" do
+        before do
+          public_record.stub(:doi).and_return(random_doi)
+        end
+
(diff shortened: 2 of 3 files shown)


spec/models/generic_file_spec.rb
```diff
@@ -625,4 +625,156 @@
       end
     end
   end
+
+  let(:public_permission_double) { double("Hydra::AccessControls::Permission") }
+  let(:gv_black_photo_sub_collection_double) { double("Collection") }
+  let(:gv_black_collection_double) { double("Collection") }
+  let(:random_doi) { "10.18131/g3-nqgv-4863" }
+  let(:blank_doi) { "" }
+  let(:public_record) { GenericFile.new(doi: []) }
```
Meowcenary, line 634: GenericFile is the system under test, so we don't really have any other options than hitting the database to create that object.

spec/models/generic_file_spec.rb
```diff
@@ -625,4 +625,156 @@
...
+  let(:private_record) { GenericFile.new(doi: []) }
+
+  # if record is publc, has a doi, and is either in the gv black photograph sub collection or is not in the greater gv black collection
```
Meowcenary, line 637: I'll go back and remove this

app/models/generic_file.rb
```diff
@@ -7,6 +7,11 @@ class GenericFile < ActiveFedora::Base
   include Galtersufia::GenericFile::FullTextIndexing
   include Galtersufia::GenericFile::MimeTypes
 
+  PUBLIC_PERMISSION = "http://projecthydra.org/ns/auth/group#public"
```
Meowcenary, line 10: I think PUBLIC_PERMISSION should absolutely be a constant, but the other two might be better off ass functions? It's probably not a big deal since either way achieves roughly the same thing.

spec/models/generic_file_spec.rb
```diff
@@ -625,4 +625,156 @@
...
+  describe "#unexportable?" do
+    before do
+      allow(public_permission_double).to receive(:agent).and_return(GenericFile::PUBLIC_PERMISSION)
```
Meowcenary, line 640: This is the meat of what I was talking about in standup. Any collaborator - something that is NOT the system under test, but used by it, should be mocked/stubbed (whatever you want to call it) so we don't need to create a ton of objects to run our test.

spec/models/generic_file_spec.rb
```diff
@@ -625,4 +625,156 @@
...
+              expect(public_record.unexportable?).to eq(false)
+            end
+          end
+        end
+      end
+    end
+
+    context "record is NOT public" do
+      before do
+        allow(private_record).to receive(:permissions).and_return([])
```
Meowcenary, line 726: I'd be willing to change `public/private_record` to just `record` since within individual context blocks we can stub out methods to have different return values. I thought that it was helpful to clarify what was going on by having two separate objects, but it does mean an extra database hit.

spec/models/generic_file_spec.rb
```diff
@@ -625,4 +625,156 @@
...
+
+  # if record is publc, has a doi, and is either in the gv black photograph sub collection or is not in the greater gv black collection
+  describe "#unexportable?" do
+    before do
+      allow(public_permission_double).to receive(:agent).and_return(GenericFile::PUBLIC_PERMISSION)
+
+      allow(gv_black_photo_sub_collection_double).to receive(:map).and_return([GenericFile::GV_BLACK_PHOTOGRAPH_SUB_COLLECTION_ID])
+      allow(gv_black_photo_sub_collection_double).to receive(:id)
+
+      allow(gv_black_collection_double).to receive(:map).and_return([GenericFile::GV_BLACK_COLLECTION_ID])
```
fenekku, line 645:
```suggestion
      allow(gv_black_collection_double).to receive(:map).and_return([GenericFile::GV_BLACK_PHOTOGRAPH_SUB_COLLECTION_ID])
```

no ?

spec/models/generic_file_spec.rb
```diff
@@ -625,4 +625,156 @@
...
+
+      context "has doi" do
+        before do
+          public_record.stub(:doi).and_return(random_doi)
+        end
+
+        context "is in the gv black photograph sub collection" do
+          context "is NOT in the gv black papers collection" do
+            before do
+              public_record.stub(:collections).and_return([gv_black_photo_sub_collection_double])
```
fenekku, line 663: Should it be 
`public_record.stub(:collections).and_return(file_collections)`?

spec/models/generic_file_spec.rb
```diff
@@ -625,4 +625,156 @@
...
+  let(:blank_doi) { "" }
+  let(:public_record) { GenericFile.new(doi: []) }
+  let(:private_record) { GenericFile.new(doi: []) }
+
+  # if record is publc, has a doi, and is either in the gv black photograph sub collection or is not in the greater gv black collection
+  describe "#unexportable?" do
+    before do
+      allow(public_permission_double).to receive(:agent).and_return(GenericFile::PUBLIC_PERMISSION)
+
+      allow(gv_black_photo_sub_collection_double).to receive(:map).and_return([GenericFile::GV_BLACK_PHOTOGRAPH_SUB_COLLECTION_ID])
```
fenekku, line 642: This is not the gv black photo sub collection, it's the collections the considered generic file is part of, no?
```suggestion
      allow(file_collections).to receive(:map).and_return([GenericFile::GV_BLACK_PHOTOGRAPH_SUB_COLLECTION_ID])
```

app/models/generic_file.rb
```diff
@@ -137,6 +142,24 @@ def export_as_chicago_citation
     add_doi_to_citation(citation)
   end
 
+  def unexportable?
+    collection_ids = collections.map(&:id)
+
+    # if record is publc, has a doi, and is either in the gv black photograph sub collection or is not in the greater gv black collection
+    public? && !doi.empty? && (collection_ids.include?(GV_BLACK_PHOTOGRAPH_SUB_COLLECTION_ID) || !collection_ids.include?(GV_BLACK_COLLECTION_ID))
+  end
```
fenekku, line 150: :warning: By this rule : A public Journal of Pediatrics paper with a DOI is not exported

`public?` true  &&
`!doi.empty?` true  &&
(
  `collection_ids.include?(GV_BLACK_PHOTOGRAPH_SUB_COLLECTION_ID)` false  ||
  !collection_ids.include?(GV_BLACK_COLLECTION_ID) true
)

So we get: true && true && true -> true for `unexportable?`which should not be the case.

Also I think we want to use the definition of "public" as being open access. For inveniordm that is currently what is being marked as [public in the export](https://github.com/galterlibrary/digital-repository/blob/5004f4dcb4ef24de9b346b641df6e910bf35d4f3/app/models/invenio_rdm_record_converter.rb#L171). @sharpattack can confirm if this is correct or not. 

**The criterion**
My understanding is that we export everything, except for 
- any open access record that doesn't have a DOI
- any record in the "old gv black collection" (see https://github.com/galterlibrary/digital-repository/issues/965). This one in particular has to be identified correctly. The metadata meeting notes + collections spreadsheet seem to be good starting points: https://northwestern.app.box.com/file/883830663475?s=lw9txtw8v2rtp4np37apg88j5qypyq9n

fenekku (commented):
The "unexportable" criterion needs to be looked at.

The mocks seem like a good idea, but hard to tell if the calls are right...

app/models/generic_file.rb
Meowcenary: Ah, nice catch. I thought there were going to be some problems with the logic TBH. It's kind of an awkward set of rules.

spec/models/generic_file_spec.rb
```diff
@@ -625,4 +625,156 @@
       end
     end
   end
+
+  let(:public_permission_double) { double("Hydra::AccessControls::Permission") }
```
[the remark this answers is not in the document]
Meowcenary: [Here is the documentation ](https://relishapp.com/rspec/rspec-mocks/docs/basics/test-doubles). As I understand it, it's an object from the Rspec library that you can send messages (function calls) to and receive pre-set return values. I think the thing that separates it from a mock or spy (they might not be a part of rspec anymore, but were in the past) is that it will verify the object you passed as a string actually has the method you specified to mock out.

spec/models/generic_file_spec.rb
Meowcenary: Good point. The way I think of things in my head does not always translate to the code, ha.

spec/models/generic_file_spec.rb
Meowcenary: I think Im misunderstanding something here. I thought it made sense for `gv_black_photo_sub_collection_double` to use the ID that ties to that same collection and similarly that the `gv_black_collection_double` should return the ID that collection. Is there any reason you want to switch these around?

spec/models/generic_file_spec.rb
Meowcenary: Yeah, this could probably be cleaned up. I was trying to make sure I covered the different test cases without creating a ton of unnecessary objects, but I think in that process I ended up adding some redundancy.

Meowcenary:
Ugh, I knew I was forgetting something with DH. I'll have to loop around to this tomorrow OR do NEC stuff. I'll check at standup.

Meowcenary:
I'm just going to close this PR and open a new one. It's difficult to read this and I think having a fresh view will be easier to make sense of.

PR #968, opened 2021-12-16
[#954] Support LCNAF Subject Name Ids

### LINKED ISSUE #954: [Export] Support LCNAF subject (name) ids
### Original problem: Non-unique subjects
e.g., Couldn't import page_gq67jr234.json --> https://digitalhub.northwestern.edu/files/gq67jr234
Error [{'subject': 'Black, G. V. (Greene Vardiman), 1836-1915.'}, {'subject': 'Stockwell, C. T.'}, {'subject': 'Black, G. V. (Greene Vardiman), 1836-1915.'}] has non-unique elements
On instance['metadata']['subjects']:
    [{'subject': 'Black, G. V. (Greene Vardiman), 1836-1915.'},
     {'subject': 'Stockwell, C. T.'},
     {'subject': 'Black, G. V. (Greene Vardiman), 1836-1915.'}]

This is typically the case when a Subject (Name) term was placed in the `subjects` field along with the same term from the Keywords (tags) DH field. This causes a duplicate error. 

#### Export side solution
Subject (Name) terms from DH should be mapped to their LCNAF id and exported like MeSH terms. For example:

`Black, G. V. (Greene Vardiman), 1836-1915` -> `"subjects": [{"id": "http://id.loc.gov/authorities/names/n80164319"}]`
 
#### Import side solution
I can do the conversion on the import side if you are blocked development-wise on the export side. In this case indicate the LCNAF terms like so:

`Black, G. V. (Greene Vardiman), 1836-1915` -> `"subjects": [{"lcnaf": "Black, G. V. (Greene Vardiman), 1836-1915"}]`

Meowcenary:
Add script to trim lcnaf.skos.ndjson file, extend existing header lookup
to include lcnaf lookup within trimmed file. closes #954

### DIFF
--- a/app/models/header_lookup.rb
+++ b/app/models/header_lookup.rb
@@ -10,18 +10,22 @@ class HeaderLookup
   MESH_ID_URI = "https://id.nlm.nih.gov/mesh/"
   MEMOIZED_MESH_FILE = "memoized_mesh.txt"
   MEMOIZED_LCSH_FILE = "memoized_lcsh.txt"
+  MEMOIZED_LCNAF_FILE = "memoized_lcnaf.txt"
   SEARCHABLE_MESH_FILE = "subjects_mesh.yml"
   SEARCHABLE_LCSH_FILE = "subjects_lcsh.yml"
+  SEARCHABLE_LCNAF_FILE = "subjects_lcnaf.csv"
 
   def initialize
     puts "initializing header_lookup..."
     # these are the terms to search through for header lookups
     @@searchable_mesh_terms ||= YAML.load_file(SEARCHABLE_MESH_FILE)
     @@searchable_lcsh_terms ||= YAML.load_file(SEARCHABLE_LCSH_FILE)
+    @@searchable_lcnaf_file ||= CSV.new(SEARCHABLE_LCNAF_FILE)
 
     # these are values that have been previously found from the searchable terms
     @@memoized_mesh ||= read_memoized_headers(MEMOIZED_MESH_FILE)
     @@memoized_lcsh ||= read_memoized_headers(MEMOIZED_LCSH_FILE)
+    @@memoized_lcnaf ||= read_memoized_headers(MEMOIZED_LCNAF_FILE)
   end
 
   def pid_lookup_by_scheme(term="", scheme="")
@@ -31,6 +35,8 @@ class HeaderLookup
       @@memoized_mesh[term] || mesh_term_pid_local_lookup(term) || nil
     elsif scheme == :lcsh
       @@memoized_lcsh[term] || lcsh_term_pid_local_lookup(term) || nil
+    elsif scheme == :subject_name
+      @@memoized_lcnaf[term] || lcnaf_pid_lookup(term) || nil
     else
       nil
     end
@@ -99,6 +105,24 @@ class HeaderLookup
     end
   end
 
+  def lcnaf_pid_lookup(lcnaf_term)
+    lcnaf_term = lcnaf_term.downcase.trim
+
+    @@searchable_lcnaf_file.each do |row|
+      term, pid = row
+
+      # if the term matches up memoize, write to file, and return pid
+      if lcnaf_term == term.downcase.trim
+        @@memoized_lcnaf[lcnaf_term] = pid
+        File.write(MEMOIZED_LCNAF_FILE, @@memoized_lcnaf)
+        return pid
+      end
+    end
+
+    # if no pid is found, return nil
+    nil
+  end
+
   private
 
   def perform_and_parse_lcsh_query(stripped_lcsh_term)
@@ -131,7 +155,7 @@ class HeaderLookup
     term.mb_chars.normalize(:kd).gsub(/[^\ \-()x00-\x7F]/n, '').to_s
   end
 
-  # read memoized from JSON file, if no file found return blank hash
+  # read memoized from file, if no file found return blank hash
   def read_memoized_headers(filepath)
     begin
       file = File.read(filepath)
--- a/spec/fixtures/mock_lcnaf.yml
+++ b/spec/fixtures/mock_lcnaf.yml
@@ -1,0 +1,2 @@
+"New York (N.Y.). Administration for Children's Services":"http://id.loc.gov/authorities/names/n2001099999"
+"Birkan, Kaarin":"http://id.loc.gov/authorities/names/n90699999"
--- a/trim_lcnaf.rb
+++ b/trim_lcnaf.rb
@@ -1,0 +1,24 @@
+require 'csv'
+require 'json'
+
+CSV.open("subjects_lcnaf.csv", "wb") do |lcnaf_csv|
+  File.open("lcnaf.skos.ndjson").each_line do |line|
+    line_hash = JSON.parse(line)
+    id = line_hash["@context"]["about"]
+    term = ""
+    graph = line_hash["@graph"]
+
+    graph.each do |graph_entry|
+      if graph_entry["@type"] == "skos:Concept"
+        term = graph_entry["skos:prefLabel"]
+      elsif graph_entry["@type"] == "skosxl:Label"
+        term = graph_entry["skosxl:literalForm"]
+      end
+
+      if !term.empty?
+        lcnaf_csv << [term, id]
+        break
+      end
+    end
+  end
+end
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -217,7 +217,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   end
 
   def resource_type(digitalhub_subtype)
-    puts "in #resource_type..."
     irdm_types = DH_IRDM_RESOURCE_TYPES[digitalhub_subtype]
 
     if irdm_types
@@ -232,7 +231,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   end
 
   def build_creator_contributor_json(creator)
-    puts "in #build_creator_contributor_json..."
     if creator_data = @@person_or_org_data[creator]
       return creator_data
     # Organization
@@ -288,7 +286,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   end # build_creator_contributor_json
 
   def format_additional(content_type, invenio_type, values)
-    puts "in #format_additional..."
     formatted_values = values.map do |value|
       if value.blank?
         next
@@ -302,7 +299,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
 
   # return array of invenio formatted subjects
   def subjects_for_scheme(terms, scheme)
-    puts "in #subjects_for_scheme..."
     mapped_terms = terms.map do |term|
       pid = @@header_lookup.pid_lookup_by_scheme(term, scheme)
 
@@ -322,7 +318,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
 
 
   def contributors(contributors)
-    puts "in #contributors..."
     contributors.map do |contributor|
       contributor_json = build_creator_contributor_json(contributor)
       contributor_json.merge!({"role": {id: ROLE_OTHER}})
@@ -331,7 +326,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   end
 
   def ark_identifiers(arks)
-    puts "in #ark_identifiers..."
     arks.map do |ark|
       {
         "identifier": ark,
@@ -341,7 +335,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   end
 
   def rights(license_urls)
-    puts "in #rights..."
     license_urls.map do |license_url|
       license_data = @@license_data[license_url.to_sym]
 
@@ -361,7 +354,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   end
 
   def version(content)
-    puts "in #version..."
     return "" unless content.has_versions?
     version_number = content.versions.all.length
 
@@ -369,7 +361,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   end
 
   def related_identifiers(related_url)
-    puts "in #related_identifiers..."
     identifiers = related_url.map do |url|
       next if url.blank?
 
@@ -394,17 +385,14 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   end
 
   def doi_url?(url)
-    puts "in #doi_url..."
     url.include?(DOI_ORG)
   end
 
   def format_publication_date(publication_date)
-    puts "in #format_publication_date..."
     normalize_date(publication_date)
   end
 
   def normalize_date(date_string)
-    puts "in #normalize_date..."
     split_date = date_string.split(/[-,\/ ]/).map(&:downcase)
     # date format starts with month first
     if (!split_date.blank? && split_date[0].length < 3)
@@ -477,7 +465,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   end
 
   def rearrange_year(date_array)
-    puts "in #rearrange_year..."
     if date_array[0].length == 4
       return date_array
     end
@@ -494,7 +481,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   end
 
   def funding(file_id)
-    puts "in #funding..."
     funding_sources = @@funding_data[file_id]
 
     if funding_sources.blank?
--- a/build_memoized_lcnaf.rb
+++ b/build_memoized_lcnaf.rb
@@ -1,0 +1,11 @@
+header_lookup = HeaderLookup.new
+
+GenericFile.all.each do |generic_file|
+  lcnaf_terms = generic_file.tag
+
+  if lcnaf_terms.blank?
+    next
+  else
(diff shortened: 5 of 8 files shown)


spec/fixtures/mock_lcnaf.yml
```diff
@@ -0,0 +1 @@
+"New York (N.Y.). Administration for Children's Services":"http://id.loc.gov/authorities/names/n2001099999"
```
Meowcenary, line 1: This is necessary because if we try to mock out File#open we would need to modify the HeaderLookup#initialize function which would ultimately be, as the kids say, a hot mess.

trim_lcnaf.rb
```diff
@@ -0,0 +1,28 @@
...
+        id = graph_entry[:"cs:subjectOfChange"][:@id]
+      end
+
+      # don't allow blanks, skip repeats
+      if previous_id != id && id != ""
+        trimmed_file << "\"#{term}\":\"#{id}\"\n"
+      end
+
+      # to avoid repeated data
+      previous_id = id
```
fenekku, line 25: Let's take this representative entry (a prettified line from the ndjson file):

```json
{
  "@context": {
    "cs": "http://purl.org/vocab/changeset/schema#",
    "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
    "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
    "rdfs1": "http://www.w3.org/1999/02/22-rdf-schema#",
    "skos": "http://www.w3.org/2004/02/skos/core#",
    "skosxl": "http://www.w3.org/2008/05/skos-xl#",
    "xsd": "http://www.w3.org/2001/XMLSchema#",
    "about": "http://id.loc.gov/authorities/names/n83166666"
  },
  "@graph": [
    {
      "@id": "http://id.loc.gov/authorities/names/n83166666",
      "@type": "skos:Concept",
      "skos:altLabel": "Huang, Yao-huang",
      "skos:changeNote": [
        {
          "@id": "_:N7fa7ac18e5ca4396963dacafeaba3535"
        },
        {
          "@id": "_:N937a169f5767452484056e1d80aec4b2"
        }
      ],
      "skos:exactMatch": {
        "@id": "http://viaf.org/viaf/sourceID/LC%7Cn++83166666#skos:Concept"
      },
      "skos:inScheme": {
        "@id": "http://id.loc.gov/authorities/names"
      },
      "skos:prefLabel": "Huang, Yaohuang",
      "skosxl:altLabel": {
        "@id": "_:Nc4f7d95d194b454298110c124741db47"
      }
    },
    {
      "@id": "_:N7fa7ac18e5ca4396963dacafeaba3535",
      "@type": "cs:ChangeSet",
      "cs:changeReason": "revised",
      "cs:createdDate": {
        "@type": "xsd:dateTime",
        "@value": "2008-12-10T07:16:30"
      },
      "cs:creatorName": {
        "@id": "http://id.loc.gov/vocabulary/organizations/ocolc"
      },
      "cs:subjectOfChange": {
        "@id": "http://id.loc.gov/authorities/names/n83166666"
      }
    },
    {
      "@id": "_:Nc4f7d95d194b454298110c124741db47",
      "@type": "skosxl:Label",
      "skosxl:literalForm": "Huang, Yao-huang"
    },
    {
      "@id": "_:N937a169f5767452484056e1d80aec4b2",
      "@type": "cs:ChangeSet",
      "cs:changeReason": "new",
      "cs:createdDate": {
        "@type": "xsd:dateTime",
        "@value": "1983-12-13T00:00:00"
      },
      "cs:creatorName": {
        "@id": "http://id.loc.gov/vocabulary/organizations/dlc"
      },
      "cs:subjectOfChange": {
        "@id": "http://id.loc.gov/authorities/names/n83166666"
      }
    }
  ]
}
``` 

- There is only 1 term per ndjson line that is identified. The `id` is always given by `line_hash["@context"]["about"]`. 
- The `"cs:ChangeSet"` type indicates there was a change, but the change is already reflected in the entry with "skos:Concept". The entries before it had a different label. Let's not worry about historical changes and just take the most recent one for now. It's a reasonable trade-off to get things done (it's actually a feature in some respect: we made all the terms in use current LCNAF terms!). 
- In my experience, I've encountered cases where `@type` was `"skosxl:Label"` (there was no `skos:Concept`) and the label (`term`) was `"skosxl:literalForm"`
- The other elements of the `"@graph"`  are skos-specific graph annotations. This should allow us to remove the "don't allow blanks/repeat part" and return immediately. This might improve the scanning speed too.
- if the prettified example was helpful, maybe a test with it is useful :smile_cat:

app/models/header_lookup.rb
```diff
@@ -31,6 +33,8 @@ def pid_lookup_by_scheme(term="", scheme="")
       @@memoized_mesh[term] || mesh_term_pid_local_lookup(term) || nil
     elsif scheme == :lcsh
       @@memoized_lcsh[term] || lcsh_term_pid_local_lookup(term) || nil
+    elsif scheme == :tag
+      @@memoized_lcnaf || lcnaf_pid_lookup(term) || nil
```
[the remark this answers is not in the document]
Meowcenary: Oh, good catch!

trim_lcnaf.rb
Meowcenary: Prettified version is very helpful. I think sometimes when I'm heads down I don't think to transform the data to something that makes more sense.

Meowcenary:
Something I thought about on the train this morning: we might want to "normalize" the terms so that lookups don't have to worry about case (upper, lower) or punctuation. I think we could "skewercaseize" the strings with something like:

- Strip all punctuation
- Replace white space with hyphens
- Downcase all the letters

I think we have something similar to this in the Galter Website project so we might be able to just copy from there.

Meowcenary:
A couple of updates on this.

To verify that the LCNAF trimming went okay I used `wc -l` on the initial JSON and compared it to the trimmed file that the script produced. I think this is good enough, but let me know if there is another way to check this.

I went to write specs for this and I realized that we would need to use a regex to normalize the strings. Initially I was just downcasing and comparing, but that was not good enough to match much. Now all the non-alphanumeric characters will be removed with spaces preserved AND it will be downcased. This was working much better in specs and I'm going to try it on staging now.

Meowcenary:
I ran this successfully on staging after making a small adjustment to the logic for exporting communities to fix a nil value that had snuck in, but might be a staging only thing. The tags did not match anything which is kind of concerning, but I'm going to run a script to see what kind of tags are on staging and if they even somewhat match up with the LCNAF file. If it seems like nonsense then I think would should merge this and try it on production.

trim_lcnaf.rb
```diff
@@ -0,0 +1,23 @@
...
+      if graph_entry[:"@type"] == "skos:Concept"
+        # set the skos:prefLabel
+        term = graph_entry[:"skos:prefLabel"]
+      elsif graph_entry[:"@type"] == "skosxl:Label"
+        term = graph_entry[:"skosxl:literalForm"]
+      end
+
+      if !term.empty?
+        trimmed_file << "\"#{term}\":\"#{id}\"\n"
+        break
```
sharpattack, line 19: We just want the first term set?

app/models/header_lookup.rb
```diff
@@ -10,18 +10,22 @@ class HeaderLookup
   MESH_ID_URI = "https://id.nlm.nih.gov/mesh/"
   MEMOIZED_MESH_FILE = "memoized_mesh.txt"
   MEMOIZED_LCSH_FILE = "memoized_lcsh.txt"
+  MEMOIZED_LCNAF_FILE = "memoized_lcnaf.txt"
   SEARCHABLE_MESH_FILE = "subjects_mesh.yml"
   SEARCHABLE_LCSH_FILE = "subjects_lcsh.yml"
+  SEARCHABLE_LCNAF_FILE = "trimmed_lcnaf.yml"
```
sharpattack, line 16: Should this be `updated_trimmed_lcnaf.yml`?

app/models/header_lookup.rb
```diff
@@ -99,6 +105,27 @@ def lcsh_term_pid_lookup(lcsh_term="")
...
+      if lcnaf_term == downcase_and_remove_non_alphanumeric(term)
+        @@memoized_lcnaf[lcnaf_term] = pid
+        File.write(MEMOIZED_LCNAF_FILE, @@memoized_lcnaf)
+        return pid
+      end
+    end
+
+    # if no pid is found, return nil
+    nil
+  end
```
sharpattack, line 127: When does `@@memoized_lcnaf` come into play? Do we want to check if the `lcnaf_term` is memoized first before checking each line of `@@searchable_lcnaf_file`?

sharpattack (commented):
Great work on this Eric. I have a couple small questions before hitting 'approve'.

app/models/header_lookup.rb
fenekku: it's here: https://github.com/galterlibrary/digital-repository/pull/968/files#diff-90953463eea6f383e933bf550db03f1b8ff5dc3e7a4c544327514c9fcdf8c410R39 (above)

app/models/header_lookup.rb
```diff
@@ -135,11 +162,17 @@ def strip_accents(term="")
   def read_memoized_headers(filepath)
     begin
       file = File.read(filepath)
+      # in read / write mode
+      # file = File.open(filepath, "w+")
```
fenekku, line 166:
```suggestion
```

trim_lcnaf.rb
```diff
@@ -0,0 +1,23 @@
+File.open("updated_trimmed_lcnaf.yml", "a") do |trimmed_file|
+  File.open("lcnaf.skos.ndjson").each_line do |line|
+    line_hash = eval(line)
```
fenekku, line 3: Isn't parsing JSON in ruby done via 

```
require 'json'
JSON.parse(line)
```
? eval executes a string as though it was Ruby code no?

trim_lcnaf.rb
```diff
@@ -0,0 +1,23 @@
...
+    graph.each do |graph_entry|
+      if graph_entry[:"@type"] == "skos:Concept"
+        # set the skos:prefLabel
+        term = graph_entry[:"skos:prefLabel"]
+      elsif graph_entry[:"@type"] == "skosxl:Label"
+        term = graph_entry[:"skosxl:literalForm"]
+      end
+
+      if !term.empty?
+        trimmed_file << "\"#{term}\":\"#{id}\"\n"
```
fenekku, line 18: If we just use a CSV writer or some such, we can skip the fiddliness above. Was performance an issue?

app/models/header_lookup.rb
```diff
@@ -99,6 +105,27 @@ def lcsh_term_pid_lookup(lcsh_term="")
...
   end
 
+  def lcnaf_pid_lookup(lcnaf_term)
+    lcnaf_term = downcase_and_remove_non_alphanumeric(lcnaf_term)
+
+    puts lcnaf_term
+    # lcnaf file is the trimmmed file here
+    @@searchable_lcnaf_file.each_line do |line|
+      # get the term and id from the file, trim off quotes and newlines
+      term, pid = line.split('":').map{ |val| val.send(:gsub, /[\"\n]/, "") }
```
fenekku, line 115: Could a CSV reader avoid this fiddliness?

app/models/header_lookup.rb
```diff
@@ -135,11 +162,17 @@ def strip_accents(term="")
...
       # eval will interpret a ruby string passed as an argument
       # in this case it constructs a hash from the file
       eval(file)
     rescue Errno::ENOENT
       return {}
     end
   end
+
+  def downcase_and_remove_non_alphanumeric(str)
+    str.downcase.gsub(/[^\w\s]/, '')
```
fenekku, line 176: Because this field is ultimately free text it might have incorrect casing, but what was the issue with non-alphanumeric? Also we can add whitespace trimming on either side.

fenekku (commented):
Apart from the comments below, I was wondering if we could generate the memoized file directly? Have you tried and found the resulting file too big to then load into memory? (it would have all the term - id pairs which _could_ be large indeed)

If that is too large:

Keeping the current approach with comments below is fine. 

OR

Wacky idea, I got in the shower: we could also do the reverse and get a much smaller memoized file. That is, we run a separate scriptA on staging/production that just gets all the LCNAF subjects we have. Then run scriptB locally to get correspondent identifiers and write to memoized file. Finally use memoized file in regular export script (maybe keep searchable file as last resort in case of added lcnaf terms in between running of scripts - not likely).

app/models/header_lookup.rb
Meowcenary: It was an attempt to try and normalize the terms that are being iterated over to find matches. I don't think it's really working though. 

After coming back to it with a clearer head I can think of a few examples where this would fail with the current replacement character. Using a space as the replacement character might work, but I think the bigger question is should we bother with trying to normalize at all?

trim_lcnaf.rb
Meowcenary: That's a good idea. I was drawing too much on how I approached the previous term lookups to work on this so I made a complicated solution a simple issue here. This is my curse, ha.

trim_lcnaf.rb
Meowcenary: I have to check again, but I think there was a reason I was not using `JSON.parse`. I think it was blowing up on the strings I was passing, but it's worth checking again as long as we're revisiting this.

app/models/header_lookup.rb
Meowcenary: Hehe, thanks

app/models/header_lookup.rb
Meowcenary: Nah, I should change it in the script that was building that. I was keeping two files at first so I could check the differences, but @fenekku's suggestions worked well so I removed the original file and renamed "updated_trimmed_lcnaf.yml" to "trimmed_lcnaf.yml"

trim_lcnaf.rb
Meowcenary: There was a conversation on the last iteration where @fenekku explained the structure of the NDJSON file. In short, each line had the term we actually wanted and then a change log of sorts and a bunch of other unnecessary stuff. I will link to it in a second.

Meowcenary: Okay, I'm having a hard time linking to the comment since it's out of date. There might be a work around, but I think the gist of it is what I was saying above.

app/models/header_lookup.rb
fenekku: Let's get a sense for how that field looks like in staging/production and then decide? I don't think we can go wrong with stripping/trimming and downcasing just for comparison sake, though e.g., "Napoleon " and "napoleon" should be treated the same.

trim_lcnaf.rb
fenekku: That's right. It's in this conversation: https://github.com/galterlibrary/digital-repository/pull/968#discussion_r770550775

carsonicator (approved):
Thanks @Meowcenary

Meowcenary:
PR is updated to use CSV reader/writer and has a small script we can use to memoize LCNAF headers without having to do a full export.

trim_lcnaf.rb
```diff
@@ -0,0 +1,23 @@
+File.open("redux_01_18_2021_trimmed_lcnaf.yml", "a") do |trimmed_file|
```
fenekku, line 1: This should be "subjects_lcnaf.csv", right?

trim_lcnaf.rb
```diff
@@ -0,0 +1,23 @@
...
+    graph.each do |graph_entry|
+      if graph_entry[:"@type"] == "skos:Concept"
+        # set the skos:prefLabel
+        term = graph_entry[:"skos:prefLabel"]
+      elsif graph_entry[:"@type"] == "skosxl:Label"
+        term = graph_entry[:"skosxl:literalForm"]
+      end
+
+      if !term.empty?
+        trimmed_file << "\"#{term}\",\"#{id}\"\n"
```
fenekku, line 18: And this would use the CSV writer no?

fenekku (commented):
I think there is a naming fix to be done and have the previous comments been looked into? We are almost there.

trim_lcnaf.rb
```diff
@@ -0,0 +1,24 @@
+require 'csv'
+require 'json'
+
```
Meowcenary, line 3: This is now using the Ruby CSV library to write out. The nice thing about this is that rows are represented as arrays so it's easy to read in values with something like: `value_a, value_b, value_c = row` which will assign value_a to row[0], value_b to row[1], and value_c to row[2].

app/models/header_lookup.rb
```diff
@@ -10,18 +10,22 @@ class HeaderLookup
...
   SEARCHABLE_MESH_FILE = "subjects_mesh.yml"
   SEARCHABLE_LCSH_FILE = "subjects_lcsh.yml"
+  SEARCHABLE_LCNAF_FILE = "subjects_lcnaf.csv"
 
   def initialize
     puts "initializing header_lookup..."
     # these are the terms to search through for header lookups
     @@searchable_mesh_terms ||= YAML.load_file(SEARCHABLE_MESH_FILE)
     @@searchable_lcsh_terms ||= YAML.load_file(SEARCHABLE_LCSH_FILE)
+    @@searchable_lcnaf_file ||= CSV.new(SEARCHABLE_LCNAF_FILE)
```
Meowcenary, line 23: This creates the CSV object, but it doesn't actually read anything right away which is great. It might be worth doing this with the other searchable files to save some time on future exports.

fenekku (approved):
Finally :)

Meowcenary:
Thanks for being patient with this!

PR #970, opened 2021-12-22
930 datacite rest api

sharpattack:
Switch to Datacite REST API from EZID API.

### DIFF
--- a/app/models/concerns/doi_generator.rb
+++ b/app/models/concerns/doi_generator.rb
@@ -1,4 +1,4 @@
-module EzidGenerator
+module DoiGenerator
   extend ActiveSupport::Concern
 
   attr_accessor :doi_message
@@ -94,7 +94,7 @@ module EzidGenerator
 
   def set_datacite_prefix(data)
     if !self.doi.present?
-      data[:data][:attributes][:prefix] = ENV['EZID_DEFAULT_SHOULDER']
+      data[:data][:attributes][:prefix] = ENV['DATACITE_DEFAULT_SHOULDER']
     end
 
     data
@@ -103,6 +103,8 @@ module EzidGenerator
   def set_datacite_event(data, hide=false)
     state = visibility_to_state
 
+    # More about events in Datacite
+    # https://support.datacite.org/docs/api-create-dois#create-a-findable-doi
     if hide
       data[:data][:attributes][:event] = "hide"
     elsif state == "findable"
@@ -114,6 +116,8 @@ module EzidGenerator
     data
   end
 
+  # Our mapping from Digitalhub visibility to Datacite state
+  # https://support.datacite.org/docs/doi-states#doi-states-outside-of-fabrica
   def visibility_to_state
     case self.visibility
     when "restricted"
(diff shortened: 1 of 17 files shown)


sharpattack:
This branch is deployed to staging. You can test uploads, see the DOI's generated, and check the test fabrica site.  https://doi.test.datacite.org/

carsonicator (approved):
Thanks @sharpattack

Meowcenary:
Thanks for handling this @sharpattack! I won't leave a review since it's already merged, but it looks good to me. Sorry for the delay on reviewing this.

app/models/concerns/doi_generator.rb
```diff
@@ -0,0 +1,170 @@
...
+        new_state = visibility_to_state
+        current_state = data["attributes"]["state"]
+        # if state changes from findable to anything else, we hide it
+        if current_state == "findable" && new_state != "findable"
+          json = datacite_api_json(hide=true)
+        else
+          json = datacite_api_json
+        end
+        client.update_metadata(identifier, json)
+        update_doi_metadata_message(current_state, new_state)
```
fenekku, line 59: It's great that a separate Datacite library wasn't needed in the end.

There might be state transition issues, so making the state transitions explicit in the code may help identify them. In any case, there is a state machine here by any other name in that code:

visibility_doi_state_machine

So something like the following to make it explicit:


```ruby
        response = client.get_doi(doi)
        data = JSON[response.body]["data"]
        identifier = data["id"]
        doi_state = data["attributes"]["state"]
        action, message = get_action_message(doi_state, self.visibility)
        json = datacite_api_json(action)
        client.update_metadata(identifier, json)
        update_doi_metadata_message(message)
```


```ruby
def get_action_message(src_doi_state, target_visiblity)
  states = {
    # (<doi_state>, <visibility>): (<action>, <message>)
  }
  return states[(src_doi_state, target_visibility)]  # add any weird exceptions here or in the states perhaps
end
```

You can then get rid of `set_datacite_event`, `visibility_to_state`....

Test what happens when a unsupported transition on Datacite REST API is executed and adjust for that. (e.g. Datacite says that "registered" or "findable" DOIs can never transition back to "draft").

app/models/concerns/doi_generator.rb
```diff
@@ -0,0 +1,170 @@
...
+    true
+  end
+
+  def create_datacite_doi
+    response = DataciteRest.new.mint(
+      datacite_api_json
+    )
+    data = JSON[response.body]["data"]
+    self.update_attributes(doi: [data["id"]])
+    self.doi_message = 'generated'
```
Meowcenary, line 38: What is `doi_message` used for? Is it something that existed in the previous DOI lookup that needs to be retained?

sharpattack: I think it is used for our purpose, not sure if it needs to be retained. I can see it being useful to check DOI states in case they are not behaving as expected.

app/models/concerns/doi_generator.rb
sharpattack: Thank you for this comment, Guillaume, this was helpful. I am glad you noticed the state machine. I knew I needed another pair of eyes after needing to just get this to work. I removed `set_datacite_event` and `visibility_to_state`, hopefully making the transitions easier to read.

PR #978, opened 2022-01-18
[#977] Debug collection_store

### LINKED ISSUE #977: Debug collection_store Creation
On Production, we get a `stack level too deep (SystemStackError)` when running export. This breaks around the collection_store creation. Likely being caused by a recursive loop(Collection belongs to itself, or two Collections jointly belonging to each other). 

Debug to output the problematic Collections, and keep the export going. Report the problematic Collections to the team, and handle appropriately.

sharpattack:
Output collections that are self containing, or is jointly contained
with another collection. Keep the export going. closes #977

### DIFF
--- a/app/models/collection_store.rb
+++ b/app/models/collection_store.rb
@@ -33,6 +33,14 @@ class CollectionStore
       starting_collection[:paths] << path
     else
       collection[:collections].each do |parent_collection|
+        if parent_collection == collection[:id]
+          puts "SELF CONTAINED COLLECTION: #{collection[:id]}"
+          next
+        elsif @data[parent_collection][:collections].include?(collection[:id])
+          puts "HAS JOINT COLLECTION: #{collection[:id]}"
+          next
+        end
+
         parent_collection_data = @data[parent_collection]
 
         recursive_collection_path(
--- a/app/models/collection_store.rb
+++ b/app/models/collection_store.rb
@@ -10,46 +10,46 @@ class CollectionStore
       @data[c.id] = {
         "id": c.id,
         "title": c.title,
-        "collections": c.collection_ids,
+        "parent_ids": c.collection_ids,
         "paths": []
       }
     end
   end
 
   def build_paths_for_collection_store
-    @data.each do |collection|
-      collection_data = collection[1]
+    @data.each do |id, collection|
       recursive_collection_path(
-        collection_data,
-        [collection_title_id_data(collection_data)],
-        collection_data
+        collection,
+        [collection_title_id_data(collection)],
+        collection
       )
     end
   end
 
   private
   def recursive_collection_path(collection, path, starting_collection)
-    if collection[:collections].empty?
+    if collection[:parent_ids].empty?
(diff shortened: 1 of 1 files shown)


sharpattack:
Going to deploy to staging to test these edge cases. Will follow up with results soon.

sharpattack:
Testing of edge cases went well. The export will finish, and output the problematic collection's id to our export log. 

FYI: the records that belong to these problematic collections will not have the `communities` info exported for now. We should address the problem as it needs to be handled regardless. Then after fixing, the collection info should be exported.

app/models/collection_store.rb
```diff
@@ -33,6 +33,14 @@ def recursive_collection_path(collection, path, starting_collection)
...
     else
       collection[:collections].each do |parent_collection|
+        if parent_collection == collection[:id]
+          puts "SELF CONTAINED COLLECTION: #{collection[:id]}"
+          next
+        elsif @data[parent_collection][:collections].include?(collection[:id])
+          puts "HAS JOINT COLLECTION: #{collection[:id]}"
+          next
+        end
+
```
fenekku, line 43: I will take a look tomorrow, but we should easily be able to just check if there is a cycle and stop the current recursion there, but continue the overall building of paths. It wouldn't even require owners to make changes to their collections.

Meowcenary: You missed an `end` after your `next` statement, but otherwise I think this is the way.

Meowcenary (commented):
I think Guillaume's suggestion is the way to go, but take a look for yourself. Also, don't forget to add the missing `end` back in if you copy that code.

app/models/collection_store.rb
sharpattack:
``` ruby
if collection_already_in_path?(parent_id, path) 
  # we've reached a cycle: don't recurse further but save the path 
  starting_collection[:paths] << path
```

We actually don't want to save the path, yet. If the loop continues, the path will get messy. We only save the path at the end of the loop. 

As I noted, we should correct the collection's relationship anyway: remove the self containing and jointly connected relationships.

fenekku: Can you expand on what you mean by the path getting messy? There are multiple paths being created.

sharpattack: After some local testing, I could not find any issues with adding the path if the collection is already there. Thanks @fenekku

app/models/collection_store.rb
```diff
@@ -10,40 +10,48 @@ def build_collection_store_data
...
+          parent_collection,
+          [collection_title_id_data(parent_collection)] + path,
           starting_collection
         )
       end
     end
   end
 
+  def collection_already_in_path?(collection_id, path)
+    path.map{|p|  p[:id]}.include?(collection_id)
```
fenekku, line 52: ah I see :curling_stone:

fenekku (approved):
Let's merge this!

carsonicator (commented):
Thanks everyone!

PR #982, opened 2022-02-14
[#125][#965] Unexportable Records

### LINKED ISSUE #125: Searching users when sharing files doesn't work

Meowcenary:
Add GenericFile#unexportable? and use it within InvenioRdmRecordConverter. closes #125, #965

### DIFF
--- a/spec/models/generic_file_spec.rb
+++ b/spec/models/generic_file_spec.rb
@@ -3,13 +3,13 @@ RSpec.describe GenericFile do
   # Tested in collection_spec
   it { is_expected.to respond_to(:add_institutional_admin_permissions) }
 
-  context 'export citations' do
-    let(:gf_doi) { GenericFile.new(title: ['abc'],
-                                   creator: ['Donald Duck'],
-                                  doi: ['doi:11111/bbbb']) }
-    let(:gf_no_doi) { GenericFile.new(title: ['meow'],
+  let(:gf_doi) { GenericFile.new(title: ['abc'],
+                                 creator: ['Donald Duck'],
+                                doi: ['doi:11111/bbbb']) }
+  let(:gf_no_doi) { GenericFile.new(title: ['meow'],
                                       creator: ['Cicero']) }
 
+  context 'export citations' do
     it 'adds a doi to apa-formated citations' do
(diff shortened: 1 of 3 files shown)


carsonicator (approved):
Thanks @Meowcenary

spec/models/generic_file_spec.rb
```diff
@@ -519,4 +519,90 @@
...
+  let(:collection_paths_gv_black_not_in_photo) { [mock_gv_black_collection] }
+  let(:collection_paths_gv_black_in_photo) { [mock_gv_black_collection, mock_gv_black_photo_collection] }
+  let(:no_gv_black) { [mock_collection] }
+
+  # if it's in the old gv black collection and not the photograph sub collection
+  describe "#in_old_gv_black_and_not_photograph" do
+    context "in old gv black collection" do
+      context "in photograph sub collection" do
+        it "returns false" do
+          expect(gf_doi.send(:in_old_gv_black_and_not_photograph?, collection_paths_gv_black_in_photo)).to eq(false)
```
fenekku, line 536: minor:

1- Although the structure `collection_paths = [ [gv_black], [gv_black_photo] ]` will pass the tests, in reality the structure is `[ [gv_black, gv_black_photo], [other_random_collection] ]` where `gv_black = {title: ..., id: ...}` and so on. That structure will also pass the tests given the code, but if one didn't know the code... 
2- Is `send` used because the method is private or is it just a Ruby idiom? Testing the method call directly might be more readable. If private, then testing the public `unexportable?` might be an idea.

None of this blocking.

fenekku (approved):

sharpattack (approved):
Great work, @Meowcenary I like the separation of the two contexts into two functions.

spec/models/generic_file_spec.rb
Meowcenary: I'll update the structure mentioned in point 1 since that seems more realistic to what we will see.

I'm using send because it is a private method. We could test `unexportable?`, but it was easier to wrap my head around testing two separate functions than one big function. When you start dropping a ton of `context` and `describe` blocks it gets very messy and hard to read. It's much cleaner the way I have it even if it's not the standard way of doing things.

PR #984, opened 2022-02-18
[#979] Export Record To Final Communities Information

### LINKED ISSUE #979: Export record to final communities information.
Each record must be tied to the community it ends up being a part of.

Meowcenary:
closes #979

### DIFF
--- a/app/models/header_lookup.rb
+++ b/app/models/header_lookup.rb
@@ -106,13 +106,13 @@ class HeaderLookup
   end
 
   def lcnaf_pid_lookup(lcnaf_term)
-    lcnaf_term = lcnaf_term.downcase.trim
+    lcnaf_term = lcnaf_term.downcase.strip
 
     @@searchable_lcnaf_file.each do |row|
       term, pid = row
 
       # if the term matches up memoize, write to file, and return pid
-      if lcnaf_term == term.downcase.trim
+      if lcnaf_term == term.downcase.strip
         @@memoized_lcnaf[lcnaf_term] = pid
         File.write(MEMOIZED_LCNAF_FILE, @@memoized_lcnaf)
         return pid
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -25,34 +25,42 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   MEMOIZED_PERSON_OR_ORG_DATA_FILE = 'memoized_person_or_org_data.txt'
   FUNDING_DATA_FILE = 'app/models/concerns/galtersufia/generic_file/funding_data.txt'
   LICENSE_DATA_FILE = 'app/models/concerns/galtersufia/generic_file/license_data.txt'
+  COLLECTIONS_TO_COMMUNITIES_JSON = 'collections_to_communities.json'
   BLANK_FUNDER_SOURCE = {funder: {name: "", identifier: "", scheme: "ror"}, award: {title: "", number: "", identifier: "", scheme: ""}}
 
   @@header_lookup ||= HeaderLookup.new
   @@funding_data ||= eval(File.read(FUNDING_DATA_FILE))
   @@person_or_org_data ||= eval(File.read(MEMOIZED_PERSON_OR_ORG_DATA_FILE))
   @@license_data ||= eval(File.read(LICENSE_DATA_FILE))
+  # parse json then trim down to a hash of dh collection id to prism community id
+  @@collections_to_communities ||= JSON.parse(File.read(COLLECTIONS_TO_COMMUNITIES_JSON)).each_with_object({}){ |node, hash| hash[node["dh"]] = node["prism"] }
 
   # Create an instance of a InvenioRdmRecordConverter converter containing all the metadata for json export
   #
   # @param [GenericFile] generic_file file to be converted for export
   def initialize(generic_file=nil, collection_store={})
-    # communites are necessary to check if the file should be exported
-    @collection_store = collection_store
-    @communities = list_collections
-
-    if generic_file&.unexportable?(@communities)
+    if generic_file.blank?
       return
     end
 
     @generic_file = generic_file
+    # communites are necessary to check if the file should be exported
+    @collection_store = collection_store
+    # communities is an array consisting of collection paths which are arrays of hashes
+    @dh_collections = list_collections
+
+    if generic_file.unexportable?(@dh_collections)
+      return
+    end
 
     @record = record_for_export
     @file = file_info
     @extras = extra_data
+    @prism_communities = map_collections_to_communities
   end
 
   def to_json(options={})
-    options[:except] ||= ["memoized_mesh", "memoized_lcsh", "generic_file", "collection_store"]
+    options[:except] ||= ["memoized_mesh", "memoized_lcsh", "generic_file", "collection_store", "dh_collections"]
     super
   end
 
@@ -214,6 +222,7 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
         "locations": {"features": @generic_file.subject_geographic.present? ? @generic_file.subject_geographic.map{ |location| {place: location.force_encoding("UTF-8")} } : []},
         "funding": funding(@generic_file.id)
       }
+
     rescue => e
       puts "[!!!ERROR!!!] Problem for GenericFile: #{@generic_file.id}"
       puts "Error - #{e}".force_encoding("UTF-8")
@@ -507,5 +516,19 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
 
     # remove nil values
     funding_sources.compact
+  end
+
+  def map_collections_to_communities
+    @dh_collections.map do |collection_path|
+      prism_community_id_from_collection_path(collection_path)
+    end
+  end
+
+  def prism_community_id_from_collection_path(collection_path)
+    collection_path.each do |collection_path_entry|
+      if community = @@collections_to_communities[collection_path_entry[:id]]
+        return community
+      end
+    end
   end
 end
--- a/spec/models/invenio_rdm_record_converter_spec.rb
+++ b/spec/models/invenio_rdm_record_converter_spec.rb
@@ -179,7 +179,7 @@ RSpec.describe InvenioRdmRecordConverter do
           "edit": [user.username]
         }
       },
-      "communities": []
+      "prism_communities": []
     }.to_json
   end
   let(:invenio_rdm_record_converter) { described_class.new(generic_file, collection_store.data) }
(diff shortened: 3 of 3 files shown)


app/models/invenio_rdm_record_converter.rb
```diff
@@ -25,34 +25,39 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
...
-    @generic_file = generic_file
-
     @record = record_for_export
     @file = file_info
     @extras = extra_data
   end
 
   def to_json(options={})
-    options[:except] ||= ["memoized_mesh", "memoized_lcsh", "generic_file", "collection_store"]
+    options[:except] ||= ["memoized_mesh", "memoized_lcsh", "generic_file", "collection_store", "communities"]
```
Meowcenary, line 60: All member variables are being exported so this would include `@communities` which is currently being assigned by `InvenioRdmRecordConverter#list_collections`: 

```
  def list_collections
    collection_paths = []

    @generic_file.collection_ids.each do |collection_id|
      @collection_store[collection_id][:paths].each do |path|
        collection_paths << path
      end
    end

    collection_paths
  end
  ```
  
  I might be wrong on this, but I thought that should be removed, then have "communities" as a field under "metadata" as I implemented in this ticket with the value of the field being the JSON objects @fenekku provided that corresponds to the files collections.

File: spec/models/invenio_rdm_record_converter_spec.rb [hunk 2]
```diff
@@ -816,4 +816,29 @@
       end
     end
   end
+
+  describe "#map_collections_to_communities" do
```
Meowcenary, line 820: Simple specs. Hopefully this will make it more clear what's going on.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -145,7 +145,8 @@
               "identifier": "",
               "scheme": ""
             }
-          }]
+          }],
+          "communities": []
```
Meowcenary, line 149: Moving this to the metadata.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -508,4 +515,19 @@ def funding(file_id)
     # remove nil values
     funding_sources.compact
   end
+
+
+  def map_collections_to_communities(communities=@communities)
```
Meowcenary, line 520: I added communities as an argument to help with testing this out. `@@collections_to_communities` is just the result of reading JSON.

Meowcenary:
The metadata documentation for InvenioRDM had nothing about communities, but I think this is the right thing to export based on previous discussion. If not, shouldn't be too hard to modify this.

spec/models/invenio_rdm_record_converter_spec.rb
fenekku: It's good where it is. Naming should be changed though.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -25,34 +25,39 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
...
   # @param [GenericFile] generic_file file to be converted for export
   def initialize(generic_file=nil, collection_store={})
+    if generic_file.blank?
+      return
+    end
+
+    @generic_file = generic_file
     # communites are necessary to check if the file should be exported
     @collection_store = collection_store
     @communities = list_collections
```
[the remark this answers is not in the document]
Meowcenary: Sure, that works for me

app/models/invenio_rdm_record_converter.rb
```diff
@@ -25,34 +25,39 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
...
   FUNDING_DATA_FILE = 'app/models/concerns/galtersufia/generic_file/funding_data.txt'
   LICENSE_DATA_FILE = 'app/models/concerns/galtersufia/generic_file/license_data.txt'
+  COLLECTIONS_TO_COMMUNITES_JSON = 'collections_to_communities.json'
   BLANK_FUNDER_SOURCE = {funder: {name: "", identifier: "", scheme: "ror"}, award: {title: "", number: "", identifier: "", scheme: ""}}
 
   @@header_lookup ||= HeaderLookup.new
   @@funding_data ||= eval(File.read(FUNDING_DATA_FILE))
   @@person_or_org_data ||= eval(File.read(MEMOIZED_PERSON_OR_ORG_DATA_FILE))
   @@license_data ||= eval(File.read(LICENSE_DATA_FILE))
+  @@collections_to_communites ||= JSON.parse(File.read(COLLECTIONS_TO_COMMUNITES_JSON))
```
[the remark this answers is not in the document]
Meowcenary: Of course.

app/models/invenio_rdm_record_converter.rb
Meowcenary: Oh, thanks! I'll go ahead and make the change

app/models/invenio_rdm_record_converter.rb
```diff
@@ -508,4 +515,19 @@ def funding(file_id)
...
+      @@collections_to_communites.map do |community_mapping_object|
+        if community[:id] == community_mapping_object["dh"]
+          community_mapping_object
+        else
+          nil
+        end
+      end
+    end
+
+    mapped_communities.flatten.compact
```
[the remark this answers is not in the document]
Meowcenary: For each collection path (first level array of `communities` here), there should be at most one Invenio community match. This means you can't flatten because you need to check each _path_ on it's own and not just the whole flattened set (whatever you want to call it).

app/models/invenio_rdm_record_converter.rb
Meowcenary: This has been updated. Marking as resolved for my own tracking.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -25,34 +25,39 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   MEMOIZED_PERSON_OR_ORG_DATA_FILE = 'memoized_person_or_org_data.txt'
   FUNDING_DATA_FILE = 'app/models/concerns/galtersufia/generic_file/funding_data.txt'
   LICENSE_DATA_FILE = 'app/models/concerns/galtersufia/generic_file/license_data.txt'
+  COLLECTIONS_TO_COMMUNITES_JSON = 'collections_to_communities.json'
```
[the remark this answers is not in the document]
Meowcenary: Updated. Resolving.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -816,4 +816,29 @@
...
+  describe "#map_collections_to_communities" do
+    let(:community_mapping_object) {
+      {
+        "title" => "Biostatistics Collaboration Center Lecture Series",
+        "dh" => "2cc92425-b656-47ea-a3b4-825405ee6088",
+        "prism" => "biostatistics-collaboration-center-lecture-series",
+        "actions" => "become"
+      }
+    }
+    let(:file_communities_with_matches){ [[{title: "Biostatistics Collaboration Center Lecture Series", id: "2cc92425-b656-47ea-a3b4-825405ee6088"}]] }
```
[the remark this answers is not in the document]
Meowcenary: There was a misunderstanding of what was meant here. Each individual collection path should be checked for a match. Each of these paths should have a single match to an Invenio community.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -178,8 +179,7 @@
           "read": ["public"],
           "edit": [user.username]
         }
-      },
-      "communities": []
+      }
     }.to_json
   end
   let(:invenio_rdm_record_converter) { described_class.new(generic_file, collection_store.data) }
```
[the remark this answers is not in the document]
Meowcenary: I think we talked about this, but the community mapping is just being read in from the class level of InvenioRdmRecordConverter. The file itself is just in the root of the app.

app/models/header_lookup.rb
```diff
@@ -106,13 +106,13 @@ def lcsh_term_pid_lookup(lcsh_term="")
   end
 
   def lcnaf_pid_lookup(lcnaf_term)
-    lcnaf_term = lcnaf_term.downcase.trim
+    lcnaf_term = lcnaf_term.downcase.strip
```
Meowcenary, line 109: I think this was something I changed in-line while building the memoized LCNAF file, but not something I changed in version control. Whoops.

Meowcenary (commented):
Updates have been made and specs are green. I'll message for re-review shortly.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -25,34 +25,42 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
...
   LICENSE_DATA_FILE = 'app/models/concerns/galtersufia/generic_file/license_data.txt'
+  COLLECTIONS_TO_COMMUNITIES_JSON = 'collections_to_communities.json'
   BLANK_FUNDER_SOURCE = {funder: {name: "", identifier: "", scheme: "ror"}, award: {title: "", number: "", identifier: "", scheme: ""}}
 
   @@header_lookup ||= HeaderLookup.new
   @@funding_data ||= eval(File.read(FUNDING_DATA_FILE))
   @@person_or_org_data ||= eval(File.read(MEMOIZED_PERSON_OR_ORG_DATA_FILE))
   @@license_data ||= eval(File.read(LICENSE_DATA_FILE))
+  # parse json then trim down to a hash of dh collection id to prism community id
+  @@collections_to_communities ||= JSON.parse(File.read(COLLECTIONS_TO_COMMUNITIES_JSON)).each_with_object({}){ |node, hash| hash[node["dh"]] = node["prism"] }
```
Meowcenary, line 36: I originally wanted to have this as a function, but I was running into trouble with the class not having access to either class or instance functions when `@@collections_to_communities` is assigned. I think this is clear enough and especially so with the comment.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -25,34 +25,42 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
...
+    if generic_file.blank?
+      return
+    end
+
+    @generic_file = generic_file
     # communites are necessary to check if the file should be exported
     @collection_store = collection_store
-    @communities = list_collections
+    # communities is an array consisting of collection paths which are arrays of hashes
+    @dh_collections = list_collections
```
Meowcenary, line 50: This name is more true to what the variable stores.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -816,4 +816,35 @@
...
     end
   end
+
+  describe "#map_collections_to_communities" do
+    let(:map_collections_to_communites_irrc) { described_class.new(generic_file, collection_store.data) }
+    let(:expected_invenio_ids) { ["biostatistics-collaboration-center-lecture-series", "center-for-community-health"] }
+
+    context "there is not a match in the collection store to the communities mapping json" do
+      before do
+        # clear the existing collections
```
Meowcenary, line 826: I figured out my issue from the previous iteration. I had declared a `before` block such that it would run before each `context` block. Database cleaner _was_ running in between each example, but then the before block was running after to re-insert the collections. Using two before blocks fixes the issue.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -508,4 +517,18 @@ def funding(file_id)
     # remove nil values
     funding_sources.compact
   end
+
+  def map_collections_to_communities
+    @dh_collections.map do |collection_path|
+      prism_community_id_from_collection_path(collection_path)
```
fenekku, line 523: This should be fine. I don't recall if we have some records that were part of a dh collection that are transferred to Prism but are not in a community there. We'll see - might filter them at import if that's the case.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -508,4 +517,18 @@ def funding(file_id)
...
+    end
+  end
+
+  def prism_community_id_from_collection_path(collection_path)
+    collection_path.each do |collection_path_entry|
+      if community = @@collections_to_communities[collection_path_entry[:id]]
```
fenekku, line 529: cute (I am assuming it's a common ruby idiom?)

(minor:
would this be clearer?
```ruby
community = @@collections_to_communities[collection_path_entry[:id]]
return community if community
```
)

fenekku (approved):
Thanks for the changes @Meowcenary . The specs read better to me as well now. (we could have even built the `collection_store.data` ourselves if hitting Solr proved to be too slow). This can be merged as-is. comments are minor.

carsonicator (approved):
Thanks everyone!

app/models/invenio_rdm_record_converter.rb
Meowcenary: I'm not sure if it's something people still use, but I've seen it used quite a bit over the years. A lot of the older ruby idioms have fallen out of favor for one reason or another. Usually it's because something could be done a different way that is considered clearer (like you've done here).

PR #988, opened 2022-03-03
replace '&' by 'and' in communities id/title

fenekku:
Honestly this is for pure synchrony: the final communities.json file is likely in galter-ir-importer or outside version control altogether.

### DIFF
--- a/communities.json
+++ b/communities.json
@@ -127,9 +127,9 @@
           }
         ]
       },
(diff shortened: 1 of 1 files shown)


sharpattack (approved):
Simple, thanks!

Meowcenary (approved):
Glad I saw this. I'll update the JSON I was working on to use this community name without the ampersand.

PR #990, opened 2022-03-08
[#902] Export Role Group

### LINKED ISSUE #902: Expand "Role Groups" into netids
We should store this information somewhere. Each "group" will have a list of user's netids(usernames).

sharpattack:
Add Role Store to export, expand users in permission group, and create
file with info of all Roles and members. closes #902

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -110,15 +110,17 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   end
 
   def file_permissions
-    permission_data = Hash.new { |h,k| h[k] = [] }
+    permission_data = Hash.new
 
     @generic_file.permissions.each do |permission|
+      permission_data[permission.access] ||= Hash.new
+
       if @role_store[permission.agent_name]
-        permission_data[permission.access].push(
-          *@role_store[permission.agent_name].keys
-        )
+        permission_data[permission.access].merge!(@role_store[permission.agent_name])
+      elsif user = User.find_by(username: permission.agent_name)
+        permission_data[permission.access].merge!({user.username => user.email})
       else
-        permission_data[permission.access] << permission.agent_name
+        permission_data[permission.access].merge!({permission.agent_name => ""})
       end
     end
 
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -87,7 +87,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
     if !@generic_file.based_near.empty?
         data["presentation_location"] = @generic_file.based_near
     end
-    data["owner"] = owner_info
     data["permissions"] = file_permissions
 
     data
@@ -97,20 +96,16 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
     user = User.find_by(username: @generic_file.depositor)
 
     if user
-      {
-        "netid": user.username,
-        "email": user.email
-      }
+      { user.username => user.email }
     else
-      {
-        "netid": "unknown",
-        "email": "unknown"
-      }
+      { "unknown": "unknown" }
     end
   end
 
   def file_permissions
     permission_data = Hash.new
+
+    permission_data["owner"] = owner_info
 
     @generic_file.permissions.each do |permission|
       permission_data[permission.access] ||= Hash.new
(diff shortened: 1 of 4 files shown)


fenekku (commented):
Thanks for this: let's also get the email while we are at it.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -112,7 +113,13 @@ def file_permissions
     permission_data = Hash.new { |h,k| h[k] = [] }
```
Meowcenary, line 113: I get that this is an older bit of code, but we should really change this to `permission_data = Hash.new` or `permission_data = {}`. Kind of verbose and distracting the way it is.

Meowcenary (approved):
This took me a bit to wrap my head around, but with the specs it was clear enough. Thanks for handling this.

app/models/role_store.rb
```diff
@@ -0,0 +1,15 @@
...
+  attr_accessor :data
+
+  def initialize
+    @data = Hash.new
+  end
+
+  def build_role_store_data
+    Role.find_each do |r|
+      @data[r.name] = {
+        "netids": r.users.map(&:username)
```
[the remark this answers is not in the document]
sharpattack: Sure, not a problem. This has been changed to: 
```
{
  "role_group_name": {
    "netid": "email"
  }
}
```

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -310,6 +311,40 @@
...
+
+    context "with role" do
+      before do
+        role_store.build_role_store_data
+      end
+
+      let(:expected_permissions) {
+        {
+          "read": ["public"],
+          "edit": ["usr1234", "exp987"]
```
fenekku, line 332: Shouldn't this be failing unless we change to

```
"edit": [
  {"netid": "usr1234", "email": <email>},
  {...}
]
```
?

sharpattack: No, this is right. The netid is the key, and that's all we are getting here. Did we want to include email in the individual record json? Email is being included in the created `tmp/export/roles_data.json`.

fenekku: After some thinking, yes I think having the email in the exported json as well would be best.

1- we already do so for the owner, so it's nice to have the same pattern.
2- would need to parse and remap the role_data.json to get a general netid to email mapping in the import

So having something like:

```
"edit": [ {"usr1234": <email>},  {...} ]
```
(same for read)

in the record would be good.

sharpattack: Okay, thank you for clarifying. The output for permissions here is now: 
```
"read": {
  "public": ""
},
"edit": {
  user.username => user.email
}
```

fenekku: I am sorry about this., I don't think I was explicit enough about the email+netid :cry:  
I meant that the generated users (from group transformation or not) should be exported with their email and netid _in all cases_. The transformation of groups into netids has highlighted how the import may receive just
some netids for users (in general), but creating a user on the import side requires an email.

We could put this in a separate task: it might be faster to do 1 query to get all netid+emails (if there is a User model) and then do the mapping everywhere. In other words, if a user was given direct access in the edit, also providing its email would be necessary. 

Second, will the difference between the colon notation and the rocket notation affect the
output? 

this:

```ruby
"owner": {
  "netid": user.username,
  "email": user.email
},
```

that:

```ruby
"edit": {
  "<netid>" => <email>
}
```

Meowcenary (approved):
Looks good to me. Thanks for iterating on this!

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -170,19 +171,24 @@
...
+          "owner": {
+            user.username => user.email
+          },
+          "read": {
+            "public": ""
+          },
+          "edit": {
+            user.username => user.email,
+          }
+        }.with_indifferent_access
```
fenekku, line 184: :heart: that's a nice structure. Really like the owner in permissions because ownership is about permission. Thanks for that @sharpattack

app/models/invenio_rdm_record_converter.rb
```diff
@@ -96,23 +96,27 @@ def owner_info
...
+
+    permission_data["owner"] = owner_info
 
     @generic_file.permissions.each do |permission|
-      permission_data[permission.access] << permission.agent_name
+      permission_data[permission.access] ||= Hash.new
+
+      if @role_store[permission.agent_name]
+        permission_data[permission.access].merge!(@role_store[permission.agent_name])
+      elsif user = User.find_by(username: permission.agent_name)
```
fenekku, line 115: If this proves too slow, we can probably preload the user in memory since there is only hundreds of them or so.

fenekku (approved):
I feel like we ended up with something better than the sum of the individual improvements here. Thanks @sharpattack !

PR #991, opened 2022-03-10
[#986] Indicate Prism Community Collections A Record Should Be A Part Of

### LINKED ISSUE #986: [Export] indicate Prism community>collection(s) a record should be part of
From recent exchange with @kglibrarian

We will eventually want to associate each record with its collection (within a community) in Prism if any. This means adding to the json export the community-collection identifiers for each record (if any). We should do it so that the export dumps have the data, because the mapping in the spreadsheet was done with respect to DH collections. In other words, once imported the mapping couldn't be found again except manually for each record which we don't want.

Because the "collections within communities" feature doesn't exist yet (and won't for a while), it's up to the implementer to decide how to encode that information in the export file. As long as the facts that a record could be part of multiple collections and a collection id is only unique within a community are respected it should be good.

(Don't have to be this way at all but information that needs to be captured should be equivalent to this 

```json
{
  # ... other fields
  "prism_collections": ["<community_id>::<collection_id>"]
}
``` 
)

- [ ] create mapping of dh collection path to Prism community id / collection id
- export that mapping on a per record basis

Meowcenary:
Add json mapping of dh collection id or file id to prism commmunity and
collection. Add function to search mapping and return string for export.
closes #986


dh_collections_prism_collection_community.json
```diff
@@ -0,0 +1,1120 @@
+{
+  "a86e1412-d72c-4cae-b8ca-16fd834cb128": {
+    "collection_id": "2019-2020",
+    "community_id": "biostatistics-collaboration-center-lecture-series"
+  },
+  "fc389d13-2430-409b-82fd-a4b26613d350": {
+    "collection_id": "Fall 2016",
```
Meowcenary, line 7: I think we might need to remap this so that collection ids are parameterized, but that should be simple. In the mapping file there were only ids for the community and not the collection so I just took the title of the collection from DH and dropped it in. If we need to remap it we could do something like:

```
JSON.parse(mapping_json).each do |entry|
  current_collection_id = entry["collection_id"]
  new_collection_id = current_collection_id.strip_special_characters.parametrize
  entry["collection_id"] = new_collection_id
end

# write out new JSON hash
```
This obviously isn't exact, but just to say it's something we can do pretty easily if need be.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -518,17 +520,36 @@ def funding(file_id)
     funding_sources.compact
   end
 
-  def map_collections_to_communities
-    @dh_collections.map do |collection_path|
-      prism_community_id_from_collection_path(collection_path)
+  def dh_collection_to_prism_community_collections
```
fenekku, line 523: Given your comment below:

```suggestion
  def dh_collection_to_prism_community_collection
```

dh_collections_prism_collection_community.json
fenekku: Passing human labels like this for the collection is totally fine. The import side will handle the translation when the time comes too. 

Also, we should be able to merge/replace `collections_to_communities.json` with this file and thus have a single source of truth for dh collection to prism community/collection (and file to community/collection). Can be done separately.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -56,7 +58,7 @@ def initialize(generic_file=nil, collection_store={})
     @record = record_for_export
     @file = file_info
     @extras = extra_data
-    @prism_communities = map_collections_to_communities
+    @prism_communities = dh_collection_to_prism_community_collections
```
fenekku, line 61: given comment below:

```suggestion
    @prism_community = dh_collection_to_prism_community_collection
```

app/models/invenio_rdm_record_converter.rb
```diff
@@ -34,6 +35,7 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   @@license_data ||= eval(File.read(LICENSE_DATA_FILE))
   # parse json then trim down to a hash of dh collection id to prism community id
   @@collections_to_communities ||= JSON.parse(File.read(COLLECTIONS_TO_COMMUNITIES_JSON)).each_with_object({}){ |node, hash| hash[node["dh"]] = node["prism"] }
+  @@dh_collection_to_prism_collection_community = JSON.parse(File.read(DH_COLLECTIONS_TO_PRISM_COLLECTION_COMMUNITY_JSON))
```
fenekku, line 38: I feel you for finding a good name here (- ω -、)ヾ(￣ω￣; )

What about one of these?

- `dh_to_prism_location`
- `dh_to_prism_hierarchy`
- `dh_to_prism_entity`
- `dh_to_prism_comlection`  (sometimes we just have to make up words: community+collection = comlection)

The reason for another name is there is a couple of things named similarly and the file doesn't just hold "collection to ... " but also "generic_file to ... " mappings. But this is more if it helps you guys.

fenekku (approved):
Thanks for that. Mapping that spreadsheet over is tedious work, but it's appreciated. 

None of the comments are blockers.

app/models/invenio_rdm_record_converter.rb
Meowcenary: Oh jeez, nice catch.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -26,6 +26,7 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   FUNDING_DATA_FILE = 'app/models/concerns/galtersufia/generic_file/funding_data.txt'
   LICENSE_DATA_FILE = 'app/models/concerns/galtersufia/generic_file/license_data.txt'
   COLLECTIONS_TO_COMMUNITIES_JSON = 'collections_to_communities.json'
```
sharpattack, line 28: Do we still need this?

sharpattack (approved):
Looks good to me. One small question if we can remove some code. As @fenekku pointed out, thank you for working with the spreadsheet!

carsonicator (approved):
Thanks @Meowcenary

PR #993, opened 2022-03-16
communities: moved to private galter-ir-importer

fenekku:
Just removing the old communities.json file.


sharpattack (approved):
Thank you! :+1:

PR #994, opened 2022-03-24
[#144] Revisit Failed Mesh Header Lookups

### LINKED ISSUE #144: Upgrade to Sufia 6.2 and Rails 4.2.x
We're already behind and need to upgrade.

Meowcenary:
Update subjects_mesh.yml to include mesh headers with qualifiers. Update
mesh lookup to format headers with qualifiers to match
subjects_mesh.yml format and return id. closes #144


Meowcenary:
Because `subjects_mesh.yml` is so much bigger now I had to use [git large file storage](https://git-lfs.github.com/) to open this PR. I think we will still need to transfer the file to staging and production with `scp` or something similar, but now we can track it in git too.

.gitattributes
```diff
@@ -0,0 +1 @@
+subjects_mesh.yml filter=lfs diff=lfs merge=lfs -text
```
Meowcenary, line 1: This is necessary for git large file storage to work.

fenekku (approved):
Thanks for that.

Meowcenary:
Re compressing and then uncompressing the file: yeah, we could, but this seems simpler :P

Meowcenary:
I'll leave this up until tomorrow afternoon in case @sharpattack or @carsonicator has anything to say. It's a pretty straightforward PR so I don't anticipate much.

sharpattack (approved):
Looks good to me. I think the wrong issue is referenced here, so be sure to close the right one after merging.

Meowcenary:
@sharpattack - good catch on the issue number. I meant to reference issue #144 in DigitalHub2 and not this repository.

PR #999, opened 2022-03-28
update dh to prism community/collection mappings [+]

fenekku:
- Updated the mapping with latest review by Karen + Josh so it matches the one in galter-ir-importer
- Removed duplicates
- Removed starting `the` (at least one of those needs to also be updated in galter-ir-importer)
- removed deprecated `collections_to_communities.json`


dh_collections_prism_collection_community.json
```diff
@@ -956,29 +953,6 @@
...
-    "community_id": "12th-general-hospital-collection"
-  },
-  "2f75r807r": {
-    "collection_id": "Portraits",
-    "community_id": "history-of-feinberg-school-of-medicine"
-  },
-  "areybook": {
-    "collection_id": "Northwestern University Medical School, 1859-1979: A pioneer in educational reform",
-    "community_id": "history-of-feinberg-school-of-medicine"
-  },
```
fenekku, line 981: Those were duplicated. Perhaps double-check if this is masking a deeper issue (some collections that should have been there?).

dh_collections_prism_collection_community.json
```diff
@@ -914,9 +914,6 @@
     "collection_id": "Northwestern University Medical School, 1859-1979: A pioneer in educational reform",
     "community_id": "history-of-feinberg-school-of-medicine"
   },
-  "e71c88e3-3a0b-4b56-90ce-8ddc75a78e81": {
-    "community_id": "pro-implementation-planning-toolkit"
-  },
```
fenekku, line 919: duplicate (see note below)

dh_collections_prism_collection_community.json
```diff
@@ -1042,9 +1016,6 @@
   "752b7c7f-ec96-4635-af07-34d4b36dac78": {
     "community_id": "history-of-feinberg-school-of-medicine"
   },
-  "3f4ac3b8-6c05-4b4d-978e-f5ab7388743b": {
-    "community_id": "history-of-feinberg-school-of-medicine"
-  },
   "9k41zd48h": {
     "community_id": "history-of-feinberg-school-of-medicine"
   },
```
fenekku, line 1021: For these `history-of-feinberg-school-of-medicine` do we have collection_ids?

sharpattack: Is this question for... the devs? @kglibrarian and Josh?

dh_collections_prism_collection_community.json
sharpattack: Same question as below.

sharpattack (approved):
Looks good to me. Approving as long as we're not waiting for changes based on the questions you posted. And thank you for removing the old json file.

dh_collections_prism_collection_community.json
fenekku: @Meowcenary committed the list, so he would know best who knows how to answer I think.

Meowcenary: @fenekku , thanks for tagging. 

These are the only two entries in `dh_collections_prism_collection_community.json` that have community_id `history-of-feinberg-school-of-medicine` and have a `collection_id`.

```
  "2f75r807r": {
    "collection_id": "Portraits",
    "community_id": "history-of-feinberg-school-of-medicine"
  },
  "areybook": {
    "collection_id": "Northwestern University Medical School, 1859-1979: A pioneer in educational reform",
    "community_id": "history-of-feinberg-school-of-medicine"
  }
```

dh_collections_prism_collection_community.json
Meowcenary: Okay, I found the duplicate entry in `dh_collections_prism_collection_community.json` (file within DigitalHub repo). I'm going to run a script to check for other duplicates too.

I'm surprised that using `JSON.parse(File.read(DH_COLLECTIONS_TO_PRISM_COLLECTION_COMMUNITY_JSON))` in the converter had issues with duplicates since it creates a hash which should not allow for duplicate keys.

Meowcenary (commented):
In addition to what we're addressing here with duplicates I separately found another entry with an ampersand while cleaning up my version of the file. I already committed that change locally, but I think you can just do a quick search for an ampersand on `dh_collections_prism_collection_community.json` to find the entry.

I figure we'll coordinate on a creating a final version of this file and then share it with each other before this is closed out.

Issue #987, opened 2022-02-25
22-02-2022-dh-export.tar.gz export issues


fenekku:
Analysis from the 22-02-2022-dh-export.tar.gz highlighted these issues to be resolved:

# 0
Pages are exported even though they are subsumed by generic files.
    Example: tmp/export/generic_file_td96k250z.json and tmp/export/page_td96k250z.json are the same
    There are 1975 pages and they are all doubles of generic_files.

# 1
Missing publication/creation date: 4
    tmp/export/generic_file_fq977t82f.json
    tmp/export/generic_file_9c67wm919.json
    tmp/export/generic_file_qb98mf520.json
    tmp/export/generic_file_b2773v70j.json
Basically: https://github.com/galterlibrary/digital-repository/issues/953

# 2
Deviant MeSH identifiers: 3
    tmp/export/generic_file_td96k250z.json
    tmp/export/generic_file_xk81jk370.json
    tmp/export/generic_file_kw52j807h.json

Of the form e.g. https://meshb.nlm.nih.gov/mesh/D000038 when they should be https://id.nlm.nih.gov/mesh/D000038
(sounds like a regression)

# 3
Duplicate subject : 150
    Example: tmp/export/generic_file_library-notes-5.json

I thought the reason for those was that some of them were unmapped LCNAF terms
i.e. one of the term comes from DH's keyword field and the other comes from DH's subject name field.
The subject name one should have been translated to its identifier, no? If this whole assumption is
not the case, deduplication should be done. (regression?)

# 4
Empty json files are exported: 1970
    Example: tmp/export/generic_file_f932d9f2-c137-4592-8af8-50447ca57647.json

I suspect those are because they were part of the records we didn't want to export, but I don't know for sure.
Either way, not exporting them at all would be clearer.

# 5
prism_communities is exported in different formats:
    tmp/export/generic_file_library-notes-5.json
    `"prism_communities": ["galter-health-sciences-library-and-learning-center"]`

    tmp/export/generic_file_cea8e826-6479-44f3-8e72-ad646e85c213.json
```
    "prism_communities": [
    [
      {
        "title": "PD language prediction experiment data and analysis scripts",
        "id": "1d4cede9-d8d6-4576-994d-91d36bd15b0b"
      }
    ]
  ]
```

Related to communities.json file having: "galter-health-sciences-library-&-learning-center"? but actually should have (as per export):
"galter-health-sciences-library-and-learning-center".

Meowcenary:
[...]

PR #1001, opened 2022-03-29
[#1000] Empty JSON Exports

### LINKED ISSUE #1000: [Export] Empty JSON Exports
See Number 4 from #987

sharpattack:
Fixes empty exported json files. Put the skip before
initializing the converter class. closes #1000

### DIFF
--- a/app/models/generic_file.rb
+++ b/app/models/generic_file.rb
@@ -159,12 +159,12 @@ class GenericFile < ActiveFedora::Base
     gfjson
   end
 
-  def unexportable?(collection_paths)
-     open_access_and_no_doi? || in_old_gv_black_and_not_photograph?(collection_paths)
+  def unexportable?
+     open_access_and_no_doi? || in_old_gv_black_and_not_photograph?
   end
 
-  def in_old_gv_black_and_not_photograph?(collection_paths)
-    in_collections?([GV_BLACK_COLLECTION_ID], collection_paths) && !in_collections?([GV_BLACK_PHOTOGRAPH_SUB_COLLECTION_ID], collection_paths)
+  def in_old_gv_black_and_not_photograph?
+    in_collections?([GV_BLACK_COLLECTION_ID]) && !in_collections?([GV_BLACK_PHOTOGRAPH_SUB_COLLECTION_ID])
   end
   private :in_old_gv_black_and_not_photograph?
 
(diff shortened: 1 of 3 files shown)


sharpattack:
Part of #987

sharpattack (commented):
I made these changes based on memory here, since I do not have the whole development environment set up. So, if is all right @Meowcenary, I would really appreciate if you are able to check a couple things for me before this gets merged in. 
- check if `#collection_ids` is a method for a GenericFile
- run the test suite
- do a test run of the export script, `repo_export_in_batches.rb` (locally or deployed to staging)

app/models/generic_file.rb
```diff
@@ -173,16 +173,8 @@ def open_access_and_no_doi?
...
-      collection_path.each do |collection_obj|
-        if collection_ids.include?(collection_obj[:id])
-          return true
-        end
-      end
-    end
-
-    false
+  def in_collections?(collection_array=[])
+    !(collection_array & self.collection_ids).empty?
```
fenekku, line 177: I don't think we can use this because being "in a collection" might mean being a couple layers down, e.g.:

```

   Collection A
   │
   └─Collection AA
     │
     └──Collection AAA
        │
        └──Record 1234
```

Record 1234 should have the call to `in_collections?([Collection_A])` return True which wouldn't be the case anymore here.

lib/scripts/repo_export_in_batches.rb
```diff
@@ -36,6 +36,8 @@
   conversion_count = 0
 
   converter[:model_class].find_each do |record_for_export|
+    next if record_for_export.unexportable?
+
     converted_record = converter[:converter_class].new(
       record_for_export, collection_store.data, role_store.data
     )
```
fenekku, line 43:
```suggestion
    converted_record = converter[:converter_class].new(
      record_for_export, collection_store.data, role_store.data
    )
    next if converted_record.skip?
```

app/models/invenio_rdm_record_converter.rb
```diff
@@ -49,10 +49,6 @@ def initialize(generic_file=nil, collection_store={}, role_store={})
     # communities is an array consisting of collection paths which are arrays of hashes
     @dh_collections = list_collections
 
-    if generic_file.unexportable?(@dh_collections)
-      return
-    end
-
```
fenekku, line 55: Extract to method `skip?`

fenekku (commented):
You've identified the right place to make the skip, it's just a matter of using what was already there instead.

app/models/generic_file.rb
sharpattack: Ahh, great catch! I completely overlooked that.

lib/scripts/repo_export_in_batches.rb
sharpattack: I would try moving the `#list_collections` collection_paths builder over to the GenericFile class. So we can still try to do the check before initializing the converter class. 

Then we can rename `#unexportable?` => `#skip?`

PR #1003, opened 2022-03-29
[#987] DigitalHub Export Fixes

### LINKED ISSUE #987: 22-02-2022-dh-export.tar.gz export issues

Meowcenary:
Addressing feedback from the 22-02-2022 export
* no longer export pages
* correct memoized mesh terms that were using the incorrect uri
* no longer export empty json
* update dh_collections_prism_collection_community.json to not use
  ampersands  - **this is going to be further addressed now that @fenekku's PR was merged with other changes to the JSON

- closes #987  (points in there that this PR doesn't address have their own issues now)


dh_collections_prism_collection_community.json
```diff
@@ -81,7 +81,7 @@
     "community_id": "researchers-collections"
   },
   "3ca02e5e-83be-4ea7-b51f-11aee3497e6c": {
-    "collection_id": "Christian Adames Publications & Presentations",
+    "collection_id": "Christian Adames Publications and Presentations",
```
fenekku, line 84: :+1: great to catch it early (the import doesn't do anything with the collection id yet, but it would probably also succumb to that issue if it was around)

lib/scripts/repo_export_in_batches.rb
```diff
@@ -9,7 +9,6 @@
 
 # set classes to have records exported and classes that will do the actual conversion
 generic_file_converter = {model_class: GenericFile, converter_class: InvenioRdmRecordConverter}
-page_converter = {model_class: Page, converter_class: InvenioRdmRecordConverter}
 collection_converter = {model_class: "", converter_class: ""}
 # TODO: Implement collection conversion
 converters = [generic_file_converter, page_converter]
```
fenekku, line 14: `page_converter` should be removed here too, right?

fenekku (approved):
Let's remove the page_converter and it's good to go.

lib/scripts/repo_export_in_batches.rb
Meowcenary: Ah, good catch.

Meowcenary:
I just checked and the this entire collection: https://digitalhub.northwestern.edu/collections/areybook , is accounted for mapping to the collection id "Northwestern University Medical School, 1859-1979: A pioneer in educational reform" and the community id "history-of-feinberg-school-of-medicine". This is the other collection that needed to be added: https://digitalhub.northwestern.edu/collections/5712m6524. I added the individual file id to prism mappings rather than just the dh collection to prism mapping since I forgot we could add by collection until I was done heh.

dh_collections_prism_collection_community.json
```diff
@@ -1087,5 +1087,29 @@
   },
   "467b71cc-a252-4a43-95fb-ca5ee8124707": {
     "community_id": "food-allergy-data-dictionary"
+  },
```
Meowcenary, line 1090: In the meeting we had this morning (29-03-2022), we noticed that these had been overlooked from the mapping JSON we created. @fenekku had other changes that have been merged and this should be the only other changes needed for now.

carsonicator (approved):
Thanks guys!

PR #1005, opened 2022-03-31
[#1002] Duplicate Subject

### LINKED ISSUE #1002: [Export] Duplicate Subject
From Number 3 in #987 :

> Duplicate subject : 150
> 
> Example: tmp/export/generic_file_library-notes-5.json
> 
> I thought the reason for those was that some of them were unmapped LCNAF terms
> i.e. one of the term comes from DH's keyword field and the other comes from DH's subject name field.
> The subject name one should have been translated to its identifier, no? If this whole assumption is
> not the case, deduplication should be done. (regression?)

Meowcenary:
Do not allow subjects that are not mapped to vocabularies to appear
twice in the export json. closes #1002

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -143,33 +143,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
       "provenance": invenio_provenance(@generic_file.proxy_depositor, @generic_file.on_behalf_of),
       "access": invenio_access(@generic_file.visibility)
     }
-    # @label = generic_file.label
-    # @depositor = generic_file.depositor
-    # @arkivo_checksum = generic_file.arkivo_checksum
-    # @relative_path = generic_file.relative_path
-    # @import_url = generic_file.import_url
-    # @resource_type = generic_file.resource_type
-    # @title = generic_file.title
-    # @creator = generic_file.creator
-    # @contributor = generic_file.contributor
-    # @description = generic_file.description
-    # @tag = generic_file.tag
-    # @rights = generic_file.rights
-    # @publisher = generic_file.publisher
-    # @date_created = generic_file.date_created
-    # @date_uploaded = generic_file.date_uploaded
-    # @date_modified = generic_file.date_modified
-    # @subject = generic_file.subject
-    # @language = generic_file.language
-    # @identifier = generic_file.identifier
-    # @based_near = generic_file.based_near
-    # @related_url = generic_file.related_url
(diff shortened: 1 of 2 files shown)


spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -878,10 +886,21 @@
       let(:subject_name_subject_type){ :subject_name }
       let(:expected_lcnaf_pid) { ["id": "http://id.loc.gov/authorities/names/n90699999"] }
 
+
       it "returns metadata for term" do
         expect(invenio_rdm_record_converter.send(:subjects_for_scheme, subject_name_terms, subject_name_subject_type)).to eq(expected_lcnaf_pid)
       end
     end
+
+    context "tag scheme" do
```
Meowcenary, line 895: This was just to help myself understand what was happening when the tag field was passed to this function.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -25,7 +26,8 @@
       creator: [user.formal_name],
       contributor: [contributor_user.formal_name],
       title: ["Primary Title"],
-      subject_name: [lcnaf_term],
+      subject_name: [lcnaf_term, duplicate_subject_term],
```
Meowcenary, line 29: As a refresher, `subject_name` terms will be checked against the LCNAF terms for a match. If there is not a match then it will return something like `{"subject": "the term without a match"}`. Terms under the `tag` field will not be checked against any vocabulary and will just output something like `{"subject": "the tag term"}`.

sharpattack (approved):
Thanks for fixing. The tests reminded me, that there can be the same LCNAF and keyword.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -143,33 +143,6 @@ def record_for_export
       "provenance": invenio_provenance(@generic_file.proxy_depositor, @generic_file.on_behalf_of),
       "access": invenio_access(@generic_file.visibility)
     }
-    # @label = generic_file.label
```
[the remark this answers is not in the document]
fenekku: :fire:

spec/models/invenio_rdm_record_converter_spec.rb
fenekku: I see this test was just for that (not for duplication testing). :+1:

fenekku (approved):
Thanks for this @Meowcenary

carsonicator (approved):
Thanks @Meowcenary

PR #1014, opened 2022-04-12
[#1012] 2022-04-8 Empty Files

### LINKED ISSUE #1012: [Export][2022-04-08] Empty files
The original 2022-04-12 export had 8047 record files (+ role_data.json).

1970 of them are files containing only `{}`. Same problem as https://github.com/galterlibrary/digital-repository/issues/1000

Examples:

generic_file_zw12z533j.json 
generic_file_zw12z534t.json 
generic_file_zw12z5353.json
generic_file_0a3ac660-009c-4369-b80e-fb1ef0e004b9.json
generic_file_0c483j37r.json
generic_file_0c483j41b.json
generic_file_1ff46015-d6de-4141-811e-1be0a8977ee8.json

| Raw export | 8047 |
|---------------|----------|
| Non `{}` files | 6077|

Meowcenary:
The script that invokes the converters for the export was checking that
the instance of a converter existed, but not if the JSON produced from
the metadata was empty or not. closes #1012

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -38,22 +38,12 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   #
   # @param [GenericFile] generic_file file to be converted for export
   def initialize(generic_file=nil, collection_store={}, role_store={})
-    if generic_file.blank?
-      puts "No file to convert"
-      return
-    end
-
     @generic_file = generic_file
     @role_store = role_store
     # communites are necessary to check if the file should be exported
     @collection_store = collection_store
     # communities is an array consisting of collection paths which are arrays of hashes
     @dh_collections = list_collections
-
-    if generic_file.unexportable?(@dh_collections)
-      puts "Record unexportable"
-      return
-    end
 
     @record = record_for_export
     @file = file_info
@@ -62,6 +52,7 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   end
 
   def to_json(options={})
+    return "{}" if @generic_file.unexportable?(@dh_collections)
     options[:except] ||= ["memoized_mesh", "memoized_lcsh", "generic_file", "collection_store", "role_store", "dh_collections"]
     super
   end
--- a/lib/scripts/repo_export_in_batches.rb
+++ b/lib/scripts/repo_export_in_batches.rb
@@ -40,9 +40,11 @@ converters.each do |converter|
     )
     puts "---------\n#{converter[:model_class].name} has id: #{record_for_export.id}\n---------"
 
-    if converted_record.present?
+    converted_record_json_string = converted_record.to_json(pretty: true)
+    converted_record_json = JSON.parse(converted_record_json_string)
+    if converted_record_json.present?
       file_path = "tmp/export/#{converter[:model_class].name.underscore}_#{record_for_export.id}.json"
-      File.write(file_path, converted_record.to_json(pretty: true))
+      File.write(file_path, converted_record_json_string)
     end
 
     conversion_count += 1
--- a/spec/models/invenio_rdm_record_converter_spec.rb
+++ b/spec/models/invenio_rdm_record_converter_spec.rb
@@ -205,14 +205,22 @@ RSpec.describe InvenioRdmRecordConverter do
   end
 
   describe "#to_json" do
-    subject { invenio_rdm_record_converter.to_json }
+    context "record is exportable" do
+      subject { invenio_rdm_record_converter.to_json }
 
-    it do
-      is_expected.to eq json
+      it "returns formatted json string" do
+        is_expected.to eq json
+      end
+    end
+
+    context "record is unexportable" do
+      it "returns blank json string" do
+        allow(generic_file).to receive(:unexportable?).and_return(true)
+        expect(invenio_rdm_record_converter.to_json).to eq "{}"
+      end
     end
   end
 
-  let(:converter) { InvenioRdmRecordConverter.new }
   let(:checksum) { "abcd1234" }
   let(:non_user_properly_formatted) { "Laster, Firston" }
   let(:personal_creator_without_user_json) {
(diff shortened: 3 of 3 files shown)


lib/scripts/repo_export_in_batches.rb
```diff
@@ -40,7 +40,8 @@
     )
     puts "---------\n#{converter[:model_class].name} has id: #{record_for_export.id}\n---------"
 
-    if converted_record.present?
+    converted_record_json = JSON.parse(converted_record.to_json)
```
Meowcenary, line 43: It's probably clear from seeing the changes, but in case it's confusing here's what was happening:

```
    converted_record = converter[:converter_class].new(
      record_for_export, collection_store.data, role_store.data
    )
```
is returning an instance of the converter class. What we really wanted to check is whether or not the resulting JSON was blank.

Meowcenary:
I'm working on the missing dates ticket as well, but I'd like to run another export with this change quickly since there were so many issues tied to this. At standup tomorrow I'll ask if we can do a quick release.

fenekku (commented):
Let's fix and test.

lib/scripts/repo_export_in_batches.rb
sharpattack: I think calling `to_json` then checking if the json is empty could work. It would need a test to confirm though. 

I think your suggested changes will work too, but with some additional notes. I will reserve them until it is decided to incorporate them.

lib/scripts/repo_export_in_batches.rb
```diff
@@ -40,7 +40,8 @@
...
+    if converted_record_json.present?
       file_path = "tmp/export/#{converter[:model_class].name.underscore}_#{record_for_export.id}.json"
       File.write(file_path, converted_record.to_json(pretty: true))
```
sharpattack, line 46: If sticking with the `converted_record_json` route, can we reuse it here, instead of calling `to_json` again? Maybe add the `(pretty: true)` to the variable?

app/models/invenio_rdm_record_converter.rb
```diff
@@ -38,30 +38,21 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
...
-    end
-
     @generic_file = generic_file
     @role_store = role_store
     # communites are necessary to check if the file should be exported
     @collection_store = collection_store
     # communities is an array consisting of collection paths which are arrays of hashes
     @dh_collections = list_collections
 
-    if generic_file.unexportable?(@dh_collections)
```
Meowcenary, line 53: Cleaning up the constructor a bit.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -205,14 +205,22 @@
...
+      subject { invenio_rdm_record_converter.to_json }
 
-    it do
-      is_expected.to eq json
+      it "returns formatted json string" do
+        is_expected.to eq json
+      end
+    end
+
+    context "record is unexportable" do
```
Meowcenary, line 216: Ensure that unexportable records return the blank json we're expecting and not something like "null"

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -269,38 +277,38 @@
   describe "#build_creator_contributor_json" do
     context 'personal record without user in digital hub, with proper name formatting' do
       it 'assigns' do
-        expect(converter.send(:build_creator_contributor_json, non_user_properly_formatted).with_indifferent_access).to eq(personal_creator_without_user_json)
+        expect(invenio_rdm_record_converter.send(:build_creator_contributor_json, non_user_properly_formatted).with_indifferent_access).to eq(personal_creator_without_user_json)
```
Meowcenary, line 280: Removing the blank file check was breaking this since there was no GenericFile object being passed in. I'm starting to wonder if at some point blank files were no longer an issue, but it remained in the logic to keep specs green.

lib/scripts/repo_export_in_batches.rb
sharpattack: Can we use the new `converted_record_json` here?

Meowcenary: Ah, missed this. I'll update it and push. Thanks for double checking me.

lib/scripts/repo_export_in_batches.rb
```diff
@@ -40,9 +40,10 @@
     )
     puts "---------\n#{converter[:model_class].name} has id: #{record_for_export.id}\n---------"
 
-    if converted_record.present?
+    converted_record_json = JSON.parse(converted_record.to_json(pretty: true))
+    if converted_record_json.present?
       file_path = "tmp/export/#{converter[:model_class].name.underscore}_#{record_for_export.id}.json"
-      File.write(file_path, converted_record.to_json(pretty: true))
+      File.write(file_path, converted_record_json)
```
fenekku, line 46: Hmm maybe this is what @sharpattack was alluding to do:

```ruby
converted_record_json_string = converted_record.to_json(pretty: true)
converted_record_json = JSON.parse(converted_record_json_string)
if converted_record_json.present?
  file_path = "tmp/export/#{converter[:model_class].name.underscore}_#{record_for_export.id}.json"
  File.write(file_path, converted_record_json_string)
```

Otherwise, doing 

```ruby
File.write(file_path, converted_record_json)
```
(as is the case right now) will write a Ruby hash to file rather than json.

(e.g.

```ruby
require 'json'

parsed = JSON.parse('{"email": null, "description": true, "o": {"o1": 2.3398}}')
File.write("dumped_json.json", parsed)
```

content of `dumped_json.json`

```
{"email"=>nil, "description"=>true, "o"=>{"o1"=>2.3398}}
```
)

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -599,11 +607,11 @@
 
   describe "#rights" do
     it 'returns the expected license information' do
-      expect(subject.send(:rights, [creative_commons_attribution_v3_url])).to eq(expected_creative_commons_attribution_v3)
-      expect(subject.send(:rights, [creative_commons_zero_url])).to eq(expected_creative_commons_zero)
-      expect(subject.send(:rights, [mit_license_url])).to eq(expected_mit)
-      expect(subject.send(:rights, [all_rights_reserved])).to eq(expected_all_rights_reserved)
-      expect(subject.send(:rights, multiple_rights)).to eq(expected_multiple_rights)
+      expect(invenio_rdm_record_converter.send(:rights, [creative_commons_attribution_v3_url])).to eq(expected_creative_commons_attribution_v3)
```
fenekku, line 610: :+1: much better naming

fenekku (commented):
another little fix :sweat_smile: ?

lib/scripts/repo_export_in_batches.rb
Meowcenary: Oh man, you're right. Whoopsie

sharpattack (approved):

carsonicator (commented):
Thanks everyone!

PR #1015, opened 2022-04-14
Release 1.26.3 Covid 19 Community And Empty File Fix

Meowcenary:
Add mapping to Covid 19 community and bug fix to prevent empty json from being exported.


sharpattack (approved):
👍

PR #1018, opened 2022-04-22
[#1016] Local Mesh Lookups Raising Errors

### LINKED ISSUE #1016: [Export] Local Mesh Lookups Raising Errors
During the staging and production export when mesh term lookups occur the below error is logged

```
GenericFile has id: kw52j808s
---------
mesh term pid local lookup
[!!!ERROR!!!] Problem for GenericFile: x633f104m
Error - undefined method `each' for #<String:0x0000000009ffa0d8>
Continuing...
```

Investigate why local lookups are failing and add specs to help debug this in the future.

Meowcenary:
Local mesh lookups are failing because git lfs is not installed on
staging or production which causes subjects_mesh.yml to not get
decompressed. Instead of using git lfs copy the subject files to each
server and add constants in local_env.rb to point to the files.
closes #1016

### DIFF
--- a/app/models/header_lookup.rb
+++ b/app/models/header_lookup.rb
@@ -11,8 +11,8 @@ class HeaderLookup
   MEMOIZED_MESH_FILE = "memoized_mesh.txt"
   MEMOIZED_LCSH_FILE = "memoized_lcsh.txt"
   MEMOIZED_LCNAF_FILE = "memoized_lcnaf.txt"
-  SEARCHABLE_MESH_FILE = "subjects_mesh.yml"
-  SEARCHABLE_LCSH_FILE = "subjects_lcsh.yml"
+  SEARCHABLE_MESH_FILE = ENV['SUBJECTS_MESH_FILE']
+  SEARCHABLE_LCSH_FILE = ENV['SUBJECTS_LCSH_FILE']
   SEARCHABLE_LCNAF_FILE = "subjects_lcnaf.csv"
 
   def initialize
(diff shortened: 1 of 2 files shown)


app/models/header_lookup.rb
```diff
@@ -11,8 +11,8 @@ class HeaderLookup
   MEMOIZED_MESH_FILE = "memoized_mesh.txt"
   MEMOIZED_LCSH_FILE = "memoized_lcsh.txt"
   MEMOIZED_LCNAF_FILE = "memoized_lcnaf.txt"
-  SEARCHABLE_MESH_FILE = "subjects_mesh.yml"
-  SEARCHABLE_LCSH_FILE = "subjects_lcsh.yml"
+  SEARCHABLE_MESH_FILE = ENV['SUBJECTS_MESH_FILE']
+  SEARCHABLE_LCSH_FILE = ENV['SUBJECTS_LCSH_FILE']
```
fenekku, line 15: minor: Is there a reason those aren't hardcoded like "memoized_mesh.txt"? 

In InvenioRDM deployments, these kinds of environment variables can come from multiple places typically split between common (not site or secret specific configurations) and site/secret specific configurations. `ENV['SUBJECTS_MESH_FILE']` doesn't seem site/secret specific.

fenekku (approved):
Just curious, this can be merged.

sharpattack (approved):

carsonicator (commented):
Thanks @Meowcenary

PR #1019, opened 2022-04-25
Release 1.26.4 Remove Searchable Subject Files From Git

Meowcenary:
The file of searchable mesh headers `subjects_mesh.yml` was previously checked into git with a plugin for large files. The plugin required additional configuration that we were unaware of and led to errors on export when the metadata included a mesh header. The mesh files are now stored directly on the servers are pointed to by environment variables that are loaded from config files.


Meowcenary:
Everything is looking good on staging.

carsonicator (commented):
Thanks @Meowcenary

PR #1020, opened 2022-04-26
[#953] Missing Dates

### LINKED ISSUE #953: [Export] Missing dates
### Current (2022-04-12)
The following four records are exported, but they are missing a valid date:

- generic_file_9c67wm919.json,The general programme of the World's Congresses of 1893 - Combined
- generic_file_qb98mf520.json,"Dinner given by the American members of the World's Columbian Dental Congress to their friends from abroad on August 18, 1893 - Combined"
- generic_file_fq977t82f.json,"Invitation to a Reception in honor of the officers and members of the World's Columbian Dental Congress held on August 12, 1893 - Combined"
- generic_file_b2773v70j.json,"Invitation to the World's Columbian Dental Congress luncheon held at the Wellington Catering Company's Cafe in the Terminal Station inside of the World's Fair Grounds on August 19, 1893 - Combined"

Their exports look like (irrelevant fields omitted):

```
{
  "record": {
    "metadata": {
      "publication_date": "", 
      "dates": [
        {
          "date": "",
          "type": {
            "id": "created"
          },
          "description": "When the item was originally created."
        }
      ],
```

Prism expects the `publication_date` to not be empty and if a date object is present for `dates`, its`date` should not be empty either.  

These records have a DOI and have passed through the GV Black filter. Should they have passed through? is a valid question.

Meowcenary:
Add wrapper function #format_dates to keep empty strings from adding
erroenous json to the export.  closes #953

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -177,10 +177,10 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
         "additional_descriptions": format_additional("description", "other", @generic_file.description.drop(1)) + format_additional("description", "acknowledgements", @generic_file.acknowledgments)\
           + format_additional("description", "abstract", @generic_file.abstract),
         "publisher": @generic_file.publisher.shift,
-        "publication_date": format_publication_date(@generic_file.date_created.shift || @generic_file.date_uploaded.to_s.force_encoding("UTF-8")),
+        "publication_date": format_publication_date(@generic_file.date_created.shift.presence || @generic_file.date_uploaded.to_s.force_encoding("UTF-8")),
         "subjects": SUBJECT_SCHEMES.map{ |subject_type| subjects_for_scheme(@generic_file.send(subject_type), subject_type) }.compact.flatten.uniq,
         "contributors": contributors(@generic_file.contributor),
-        "dates": @generic_file.date_created.map{ |date| {"date": normalize_date(date), "type": {"id": "created"}, "description": "When the item was originally created."} },
+        "dates": format_dates(@generic_file.date_created),
         "languages": @generic_file.language.map{ |lang| lang.present? && lang.downcase == ENGLISH ? {"id": "eng"} : nil }.compact,
         "identifiers": ark_identifiers(@generic_file.ark),
         "related_identifiers": related_identifiers(@generic_file.related_url),
@@ -373,6 +373,16 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
 
   def format_publication_date(publication_date)
     normalize_date(publication_date)
+  end
+
+  def format_dates(dates)
+    dates.reject(&:blank?).map do |date|
+      {
+        "date": normalize_date(date),
+        "type": {"id": "created"},
+        "description": "When the item was originally created."
+      }
+    end
   end
 
   def normalize_date(date_string)
--- a/spec/models/invenio_rdm_record_converter_spec.rb
+++ b/spec/models/invenio_rdm_record_converter_spec.rb
@@ -709,6 +709,14 @@ RSpec.describe InvenioRdmRecordConverter do
     end
   end
 
+  describe "#format_dates" do
+    context "file has empty string for dated created" do
+      it "returns empty array" do
+        expect(invenio_rdm_record_converter.send(:format_dates, [""])).to eq([])
+      end
+    end
+  end
+
   describe "#normalize_date" do
     context "formatted date" do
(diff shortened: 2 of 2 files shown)


app/models/invenio_rdm_record_converter.rb
```diff
@@ -177,11 +177,11 @@ def invenio_metadata
...
           + format_additional("description", "abstract", @generic_file.abstract),
         "publisher": @generic_file.publisher.shift,
-        "publication_date": format_publication_date(@generic_file.date_created.shift || @generic_file.date_uploaded.to_s.force_encoding("UTF-8")),
+        "publication_date": format_publication_date(@generic_file.date_created.shift.presence || @generic_file.date_uploaded.to_s.force_encoding("UTF-8")),
         "subjects": SUBJECT_SCHEMES.map{ |subject_type| subjects_for_scheme(@generic_file.send(subject_type), subject_type) }.compact.flatten.uniq,
         "contributors": contributors(@generic_file.contributor),
-        "dates": @generic_file.date_created.map{ |date| {"date": normalize_date(date), "type": {"id": "created"}, "description": "When the item was originally created."} },
-        "languages": @generic_file.language.map{ |lang| lang.present? && lang.downcase == ENGLISH ? {"id": "eng"} : nil }.compact,
+        "dates": format_dates(@generic_file.date_created),
+        "language": @generic_file.language.map{ |lang| lang.present? && lang.downcase == ENGLISH ? {"id": "eng"} : nil }.compact,
```
Meowcenary, line 184: This was breaking specs in a way that was very annoying to track down >_<

Meowcenary: Is this supposed to be "languages" or just singular "language"? I don't remember changing this line so it might have come from a rebase or something, but then I'm left wondering why it's changed here...

app/models/invenio_rdm_record_converter.rb
```diff
@@ -375,6 +375,15 @@ def format_publication_date(publication_date)
     normalize_date(publication_date)
   end
 
+  def format_dates(dates)
```
Meowcenary, line 378: This seemed like the easiest way to add the logic we needed. There might be a way to make it a one-liner with the `{}` block syntax, but if there is it will be very, very ugly. This may not be super good looking, but it gets the point across.

app/models/invenio_rdm_record_converter.rb
fenekku: It should be `languages`

app/models/invenio_rdm_record_converter.rb
```diff
@@ -177,11 +177,11 @@ def invenio_metadata
         "additional_descriptions": format_additional("description", "other", @generic_file.description.drop(1)) + format_additional("description", "acknowledgements", @generic_file.acknowledgments)\
           + format_additional("description", "abstract", @generic_file.abstract),
         "publisher": @generic_file.publisher.shift,
-        "publication_date": format_publication_date(@generic_file.date_created.shift || @generic_file.date_uploaded.to_s.force_encoding("UTF-8")),
+        "publication_date": format_publication_date(@generic_file.date_created.shift.presence || @generic_file.date_uploaded.to_s.force_encoding("UTF-8")),
```
fenekku, line 180: Oh wow, I see it's because Ruby doesn't treat `[]` or `""` as falsy. Got it.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -375,6 +375,15 @@ def format_publication_date(publication_date)
     normalize_date(publication_date)
   end
 
+  def format_dates(dates)
+    formatted_dates = dates.map do |date|
+        next if date.blank?
+        {"date": normalize_date(date), "type": {"id": "created"}, "description": "When the item was originally created."}
```
fenekku, line 381: minor: What about smaller line lengths + filter-map pattern? So something like

```ruby
def format_dates(dates)
  dates.select(&:blank?).map do |date|
    {
      "date": normalize_date(date), 
      "type": {"id": "created"}, 
      "description": "When the item was originally created."
    }
  end
end
```

fenekku (commented):
Yup it needs to be `languages` (minor recommendation too but languages is the real thing).

Somewhat related note, I found https://rubystyle.guide/ that seems a good resource for style guidance.

app/models/invenio_rdm_record_converter.rb
Meowcenary: I'm fine with smaller line lengths. At some point it might be nice to add a style linter. I actually kind of like `next if` on it's own line, but this is better.

fenekku (approved):
Thanks!

PR #1034, opened 2022-06-24
Release 1.26.5 Identify Unmappable Subjects And Missing Dates Bug

Meowcenary:
Script to identify terms that are unmappable and minor bug fix for missing dates during export.


fenekku (approved):

Meowcenary:
I'll wait a bit longer in case anyone wants to look this over. I'll plan on merging and deploying around 3pm central

carsonicator (approved):
Thanks @Meowcenary

Meowcenary:
Decided to hold off on merging for Monday since the day is already over :( There will also be a website release so it should be easy enough to remember.

PR #1038, opened 2022-07-01
[#1036] Update LCSH MeSH Terms Lookup

### LINKED ISSUE #1036: [Export] Update LCSH/MeSH terms lookup
Analysis on Prism of missed LCSH / MeSH terms:

MeSH terms (with qualifier replacement - see below)
Found 982 valid terms out of 1144 missed terms

LCSH terms
Found 274 valid terms out of 318 missed terms

[lcsh_analysis.csv](https://github.com/galterlibrary/digital-repository/files/9003971/lcsh_analysis.csv)
[mesh_analysis.csv](https://github.com/galterlibrary/digital-repository/files/9003972/mesh_analysis.csv)

Task is to update the lookup code to:

- use the [subjects_mesh.jsonl](https://github.com/galterlibrary/invenio-subjects-mesh/blob/master/invenio_subjects_mesh/vocabularies/subjects_mesh.jsonl) and [subjects_lcsh.jsonl](https://github.com/galterlibrary/invenio-subjects-lcsh/blob/master/invenio_subjects_lcsh/vocabularies/subjects_lcsh.jsonl) as source of lookup
- make sure to perform qualifier replacement (i.e. replace "--" by "/") for **MeSH terms only**
- don't bother going out to the network

The unfound terms after this are minimal and I think we can ignore them.

Meowcenary:
Switch LCSH and MeSH searchable files to jsonl format and use lazy
loading to avoid memory usage issues. Remove old subject header lookups
that were using network calls - everything is through local files now

Reset cached lookups so they can be recreated with new files

closes #1036


memoized_mesh.txt
```diff
@@ -1 +1 @@
...
+{"Vocabulary, Controlled"=>"https://id.nlm.nih.gov/mesh/D018875", "Bile Duct Neoplasms"=>"https://id.nlm.nih.gov/mesh/D001650", "Burkitt Lymphoma--etiology"=>"https://id.nlm.nih.gov/mesh/D002051Q000209"}
```
Meowcenary, line 1: What's here is from the specs. At some point I need to revisit mocking out the memoized read and memoized write calls, but for now I'm going to leave them alone since will unblock other work.

memoized_lcsh.txt
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+{"Semantic Web"=>"http://id.loc.gov/authorities/subjects/sh2002000569", "Abdomen--Cancer"=>"http://id.loc.gov/authorities/subjects/sh85000095"}
```
Meowcenary, line 1: This is from specs. In order to get better matches this will need to be rebuilt

app/models/header_lookup.rb
```diff
@@ -70,41 +68,6 @@ def lcsh_term_pid_local_lookup(lcsh_term="")
     nil
   end
 
-  # return PID for provided mesh_header using SPARQL query
```
Meowcenary, line 73: No more networked lookups!

app/models/header_lookup.rb
```diff
@@ -1,25 +1,21 @@
 class HeaderLookup
-  BASE_SPARQL_MESH_URI ="https://id.nlm.nih.gov/mesh/sparql?format=JSON&limit=10&inference=true&query=PREFIX%20rdfs"\
```
Meowcenary, line 2: No longer needed!

app/models/header_lookup.rb
```diff
@@ -1,25 +1,21 @@
...
+  SEARCHABLE_MESH_FILE = 'subjects_mesh.jsonl'
+  SEARCHABLE_LCSH_FILE = 'subjects_lcsh.jsonl'
 
   def initialize
     puts "initializing header_lookup..."
     # these are the terms to search through for header lookups
-    @@searchable_mesh_terms ||= YAML.load_file(SEARCHABLE_MESH_FILE)
-    @@searchable_lcsh_terms ||= YAML.load_file(SEARCHABLE_LCSH_FILE)
+    @@searchable_mesh_file ||= SEARCHABLE_MESH_FILE
+    @@searchable_lcsh_terms ||= SEARCHABLE_LCSH_FILE
```
fenekku, line 18: minor: Typo on consistent name (both end in `_file` or both end in `_terms`)

Meowcenary: It's kind of concerning that this was still passing specs, but I guess that just means it was a consistent misspelling heh.

PR #1039, opened 2022-07-07
[#1032] Geographic Subject Terms

### LINKED ISSUE #1032: [Export/Import] Geographic subject terms
Those were placed in `"metadata.locations.features": [{"place": <here>}]` on Prism, but it makes more sense to place them in `metadata.subjects` under a vocabulary.

We need to see what vocabulary was used on DH for this field. Can we re-use it in Prism? If yes export to this field.
e.g., if it's LCNAF, then we can export to LCNAF

According to https://github.com/galterlibrary/digital-repository/issues/1033#issuecomment-1169107264, the vocabulary used in DH was a subset of LCNAF (only heading and only direct geographic locations)

Meowcenary:
Update export to search lcnaf file for geographic subject names. Update
build_memoized_lcnaf.rb to search lcnaf for subject_name and
subject_geographic fields to build cache. Update lcnaf searchable file
to use jsonl instead of csv. closes #1032

### DIFF
--- a/app/models/header_lookup.rb
+++ b/app/models/header_lookup.rb
@@ -7,16 +7,17 @@ class HeaderLookup
   MEMOIZED_LCSH_FILE = "memoized_lcsh.txt"
   MEMOIZED_LCNAF_FILE = "memoized_lcnaf.txt"
 
-  SEARCHABLE_LCNAF_FILE = "subjects_lcnaf.csv"
+  SEARCHABLE_LCNAF_FILE = "subjects_lcnaf.jsonl"
   SEARCHABLE_MESH_FILE = 'subjects_mesh.jsonl'
   SEARCHABLE_LCSH_FILE = 'subjects_lcsh.jsonl'
 
+  ABSENT_SUBJECT = :absent
+
   def initialize
-    puts "initializing header_lookup..."
     # these are the terms to search through for header lookups
     @@searchable_mesh_file ||= SEARCHABLE_MESH_FILE
     @@searchable_lcsh_file ||= SEARCHABLE_LCSH_FILE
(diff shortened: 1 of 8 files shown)


Meowcenary:
I will need to move the searchable LCNAF jsonl file to the server with scp or something similar. It's much too big for git.

convert_lcnaf_csv_to_jsonl.rb
```diff
@@ -0,0 +1,24 @@
...
+        term = graph_entry["skosxl:literalForm"]
+      end
+
+      if !term.empty?
+        jsonl_file << {"id": id.to_s, "scheme": "LCNAF", "subject": term.to_s}.to_json + "\n"
+        break
+      end
+    end
+  end
+end
```
fenekku, line 24: This is converting an `.ndjson` into `jsonl`, right? (No csv)

I would guess this helps a little since each json is not as big per line as the original ndjson..., but csv would then save more memory on top of that. Using csv might have been fine and take less space. 

I am sorry I probably biased us in favor of .jsonl files. I originally made the choice in these other contexts, because there might have been added nested fields. CSV doesn't give us nested objects per line nicely as opposed to jsonl files. (in my case it might even have been premature looking back on it - vocabularies might introduce nested translated titles, but it's a might...)

fenekku (commented):
Couple of comments to help us take advantage of the cache fully.

convert_lcnaf_csv_to_jsonl.rb
Meowcenary: Oh jeez - thanks. I just copied the original file and quickly changed things without intending to include it at first, but I figured it would be worth having on here. Even better to have it with a good name and no unnecessary libraries.

fenekku (approved):
Excellent, thanks for this.

PR #1040, opened 2022-07-07
[#1036] 1036-update_lcsh_mesh_terms_lookup

### LINKED ISSUE #1036: [Export] Update LCSH/MeSH terms lookup

Meowcenary:
Strip whitespace from terms before lookup. closes #1036

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -284,6 +284,7 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   # return array of invenio formatted subjects
   def subjects_for_scheme(terms, scheme)
     mapped_terms = terms.map do |term|
+      term = term.strip
       pid = @@header_lookup.pid_lookup_by_scheme(term, scheme)
 
       if term.blank?


fenekku (approved):
to the point

Meowcenary:
@gneidhardt

For point 1, if i'm understanding right the term has been updated in digitalhub so I think this should be resolved on the next export. If the term is not in the current LCSH file that we have we can add an entry to the memoized LCSH header file to resolve this. I don't think we need to re-open the issue for that since it's a small change that can be added with something else if it is needed.

For point 2 we can resolve this by copying changes we made for LCNAF memoization to LCSH and MeSH so that both the successful searches **and** the failed searches are memoized. That way after the next export is run we can generate a list of failed terms and map as needed. I'll enter a separate ticket for this.

fenekku:
I moved Gretchen's comment to its own issue and summarized the action points there: https://github.com/galterlibrary/digital-repository/issues/1042

PR #1043, opened 2022-07-11
[#1033] Export Location

### LINKED ISSUE #1033: [Export] Location
*[Edited 2022-06-30 to take into account [comment 1](https://github.com/galterlibrary/digital-repository/issues/1033#issuecomment-1169107264)]*

In DH, geographic subject names are used to describe the locations the record is about. And location is used to describe the geographical location of presentation.

For the task about exporting geographic subject names go here: https://github.com/galterlibrary/digital-repository/issues/1032 .

This task is about exporting the location. 

| DH field                                                          | Prism				 |
| ----------------------------------------------------------------- | ------------------ |
| `location` (double-check) | presentation_location |

To export the original presentation location, add an entry to the `additional_descriptions` field in the export. The added additional_description item will look like:

```json
    {
      "description": "presentation_location: <value>",
      "type": {
        "id": "other"
      }
    },
```

Make sure locations are properly exported e.g. https://digitalhub.northwestern.edu/files/nv9352841 didn't have "Tucson, Arizona, United States" exported at all in the past.

This overrides how we were exporting locations before. We don't use the [InvenioRDM locations](https://inveniordm.docs.cern.ch/reference/metadata/#locations-0-n) field anymore for now, because it has the same meaning as the subjects field we will use.

Meowcenary:
Export the GenericFile filed based_near values with the prefix
"presentation_location: " as additional descriptions with the type
"other". closes #1033

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -77,7 +77,7 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
     data = {}
 
     if !@generic_file.based_near.empty?
-        data["presentation_location"] = @generic_file.based_near
+      data["presentation_location"] = @generic_file.based_near
     end
     data["permissions"] = file_permissions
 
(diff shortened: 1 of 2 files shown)


carsonicator (approved):
Thanks @Meowcenary

Meowcenary:
> Looks good. Were you able to determine the issue with https://digitalhub.northwestern.edu/files/nv9352841 ?

I wasn't able to replicate the issue locally so I'll have to check it out on a terminal after the next release.

PR #1045, opened 2022-07-12
[#1041] Memoize Failed Lookups For Mesh and Lcsh

### LINKED ISSUE #1041: [Export] Memoize Failed Lookups For MeSH and LCSH
Update mesh and lcsh header lookup to memoize failed lookups with the key-value pair `<failed lookup term>: :absent`. These will be almost the exact same as the LCNAF lookup function.

Meowcenary:
Update header lookup to Memoize failed lookups for mesh and lcsh terms.
Remove unnecessary code from header lookup, converter, and respective
specs. closes #1041


app/models/header_lookup.rb
```diff
@@ -1,15 +1,11 @@
 class HeaderLookup
-  LCSH_BASE_URI = "http://id.loc.gov/authorities/subjects/suggest/?q="
```
Meowcenary, line 2: These should have been removed before, but I think they were kept in to keep some specs working.

app/models/header_lookup.rb
```diff
@@ -1,15 +1,11 @@
...
-  MESH_ID_URI = "https://id.nlm.nih.gov/mesh/"
-
   MEMOIZED_MESH_FILE = "memoized_mesh.txt"
   MEMOIZED_LCSH_FILE = "memoized_lcsh.txt"
   MEMOIZED_LCNAF_FILE = "memoized_lcnaf.txt"
 
-  SEARCHABLE_LCNAF_FILE = "subjects_lcnaf.jsonl"
   SEARCHABLE_MESH_FILE = 'subjects_mesh.jsonl'
   SEARCHABLE_LCSH_FILE = 'subjects_lcsh.jsonl'
+  SEARCHABLE_LCNAF_FILE = "subjects_lcnaf.jsonl"
```
Meowcenary, line 8: I thought it would be nice to keep the order consistently mesh, lcsh, lcnaf

app/models/header_lookup.rb
```diff
@@ -26,70 +22,53 @@ def initialize
...
+    elsif field == :subject_name || field == :subject_geographic || field = :lcnaf
+      searchable_file = @@searchable_lcnaf_file
+      memoized_terms = @@memoized_lcnaf
+      memoized_file = MEMOIZED_LCNAF_FILE
     else
-      nil
-    end
-  end
-
-  def mesh_term_pid_local_lookup(mesh_term="")
```
Meowcenary, line 48: I've been wanting to refactor all the lookup functions into something more generic for awhile now. It seemed like this would be a good opportunity.

app/models/header_lookup.rb
```diff
@@ -26,70 +22,53 @@ def initialize
...
     end
 
-    @@memoized_lcnaf[lcnaf_term] = lcnaf_id
-    File.write(MEMOIZED_LCNAF_FILE, @@memoized_lcnaf)
-    lcnaf_id
+    if term_pid == ABSENT_SUBJECT
+      nil
+    else
+      term_pid
+    end
```
fenekku, line 71: minor: We can just go for the one liner here to grok the gist of it quicker: 

 ```suggestion
    if term_pid == ABSENT_SUBJECT then nil else term_pid end
```

File: app/models/header_lookup.rb [hunk 5]
```diff
@@ -26,70 +22,53 @@ def initialize
...
-        return mesh_id
-      end
+      return nil
     end
 
-    nil
-  end
+    # check for memoized term
+    if term_pid = memoized_terms[term]
+      term_pid
```
fenekku, line 49: minor: having `=` in a conditional is needlessly worrisome
 
```suggestion
    if memoized_terms[term]
      term_pid = memoized_terms[term]
```

app/models/invenio_rdm_record_converter.rb
```diff
@@ -288,15 +288,11 @@ def subjects_for_field(terms, field)
...
-      pid = @@header_lookup.pid_lookup_by_field(term, field)
-
-      if pid.present?
-        {id: pid}
-      elsif field == :subject_name || field == :tag
+        nil
+      elsif field == :tag
         {subject: term.force_encoding("UTF-8")}
+      elsif pid = @@header_lookup.pid_lookup_by_field(term, field)
+        {id: pid}
```
fenekku, line 295: So we don't let unfound terms through now? Works for me, let's pay attention if librarians say anything after new export.

Also, I think this is actually a good exception to typically not placing `=` in conditionals: an intensive operation included in larger case branching that would have broken it up otherwise or required expensive recomputation :heavy_check_mark: .

spec/models/header_lookup_spec.rb
```diff
@@ -2,78 +2,45 @@
...
-
   let(:mesh_local_subject) { "Abrin" }
   let(:expected_mesh_local_subject) { "https://id.nlm.nih.gov/mesh/D000036" }
 
-  describe "#mesh_term_pid_local_lookup" do
-    it "returns expected pid" do
-      expect(subject.mesh_term_pid_local_lookup(mesh_local_subject)).to eq(expected_mesh_local_subject)
-    end
-  end
-
```
fenekku, line 70: :fire:

spec/models/header_lookup_spec.rb
```diff
@@ -2,78 +2,45 @@
...
   let(:lcnaf_term) { "Birkan, Kaarin" }
   let(:lcnaf_id) { "http://id.loc.gov/authorities/names/n90699999" }
 
-  describe "#lcnaf_pid_lookup" do
-    it "it returns the correct id for the lcnaf term" do
-      expect(subject.lcnaf_pid_lookup(lcnaf_term)).to eq(lcnaf_id)
+  describe "#pid_lookup_by_field" do
+    context "nil values" do
+      it 'returns nil for subject header with no pid' do
```
fenekku, line 19: minor: to distinguish them when errors and reveal what they are getting at

```suggestion
      it 'returns nil for a subject header with no pid' do
```

spec/models/header_lookup_spec.rb
```diff
@@ -2,78 +2,45 @@
...
+        expect(subject.send(:pid_lookup_by_field, no_pid_lcsh_subject, :lcsh)).to eq(nil)
+      end
+
+      it 'returns nil for subject header with no pid' do
```
fenekku, line 23:
```suggestion
      it 'returns nil for another subject header with no pid' do
```

spec/models/header_lookup_spec.rb
```diff
@@ -2,78 +2,45 @@
...
+        expect(subject.send(:pid_lookup_by_field, no_pid_mesh_subject, :mesh)).to eq(nil)
```
fenekku, line 24: minor: the syntax used below without `send` seems more readable to me / representative of usage. Could use it here? Maybe it's a ruby cultural thing too.

spec/models/header_lookup_spec.rb
```diff
@@ -2,78 +2,45 @@
...
+
+    context "#mesh_term_pid_local_lookup" do
+      it "returns expected pid" do
+        expect(subject.pid_lookup_by_field(mesh_local_subject, :mesh)).to eq(expected_mesh_local_subject)
+      end
+    end
+
+    context "#lcnaf_pid_lookup" do
+      it "it returns the correct id for the lcnaf term" do
+        expect(subject.pid_lookup_by_field(lcnaf_term, :lcnaf)).to eq(lcnaf_id)
```
fenekku, line 42: This is the same kind of test as the 2 above, right? If so parallel naming structure would be clearer.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -920,7 +920,7 @@
     context "subject_name scheme" do
       let(:subject_name_term){ ["malignant"] }
       let(:subject_name_subject_type){ :subject_name }
-      let(:expected_subject_name_result){ ["subject": "malignant"] }
+      let(:expected_subject_name_result){ [] }
```
fenekku, line 923: :ballot_box_with_check:  from change above

fenekku (approved):

PR #1050, opened 2022-07-25
Release 1.28.0 July Export Updates

Meowcenary:
- [Memoize failed header lookups](https://github.com/galterlibrary/digital-repository/issues/1041)
- [Support exporting languages other than English](https://github.com/galterlibrary/digital-repository/issues/1031)
- [Export location field](https://github.com/galterlibrary/digital-repository/issues/1033)
- [Update mesh/lcsh terms to use JSONL](https://github.com/galterlibrary/digital-repository/issues/1036)
- [Export geographic subject terms](https://github.com/galterlibrary/digital-repository/issues/1032)


Meowcenary:
oh no, merge conflicts :(

app/models/header_lookup.rb
```diff
@@ -1,15 +1,12 @@
...
   MEMOIZED_LCSH_FILE = "memoized_lcsh.txt"
   MEMOIZED_LCNAF_FILE = "memoized_lcnaf.txt"
 
-  SEARCHABLE_LCNAF_FILE = "subjects_lcnaf.jsonl"
-  SEARCHABLE_MESH_FILE = 'subjects_mesh.jsonl'
-  SEARCHABLE_LCSH_FILE = 'subjects_lcsh.jsonl'
+  SEARCHABLE_MESH_FILE = "subjects_mesh.jsonl"
+  SEARCHABLE_LCSH_FILE = "subjects_lcsh.jsonl"
+  # Needs to be in config because file is too large for VCS
+  SEARCHABLE_LCNAF_FILE = ENV["SUBJECTS_LCNAF"]
```
Meowcenary, line 9: LCNAF is just too big to include with the repo :( Frustrating, but it will work.

Meowcenary:
Going to put this on staging and give it a little test. If we end up with some more memoized results it might be a good chance to add those to this release too.

app/models/header_lookup.rb
sharpattack: So an LCNAF file will be copied onto the server?

sharpattack (approved):
Release PR, it looks good to me. What was the merge conflict? 

Would it be a good idea to run the test suite locally and report back? I think some are expected to fail anyway, but it would be nice to check if it's something new or not.

Meowcenary:
> Release PR, it looks good to me. What was the merge conflict?
> 
> Would it be a good idea to run the test suite locally and report back? I think some are expected to fail anyway, but it would be nice to check if it's something new or not.

I've been running the spec for the exporter before every merge, but I can run it again. I don't really run the larger test suite because i'm not changing any of that behavior and it's just such a mess. Lots of specs have been failing since I first cloned it and it's very slow.

sharpattack:
Yea, I did the same, too. After a while, I think it's a good idea to run the whole suite. 

Got it, just checking that production will have the same LCNAF file too. 

As long as it will not impact the `master` branch for future merging, fine by me.

fenekku (approved):
:shipit:

Meowcenary:
> Test Suite: Yea, I did the same, too. After a while, I think it's a good idea to run the whole suite.
.
Fair point. I'll run the whole suite locally since we haven't released in awhile and if that's all good I'll merge. I'll plan on exporting overnight so we don't get in the way of anyone who might be using DH

PR #1051, opened 2022-07-28
[#1048] Last Editor Info

### LINKED ISSUE #1048: Add netid of last editor to Subjects.csv

sharpattack:
For subjects.csv metadata export, include last peformed event for
record, and collection ids fix. DRY refactor. closes #1048

### DIFF
--- a/lib/batch/subject.csv.rb
+++ b/lib/batch/subject.csv.rb
@@ -1,26 +1,32 @@
 csv = CSV.open('/home/deploy/all_gf_subject.csv', 'wb')
 terms =  GalterGenericFilePresenter.terms + [
-  :depositor, :date_uploaded, :date_modified, :read_access_group, :isPartOf, :id
+  :depositor, :date_uploaded, :date_modified, :read_access_group, :collection_ids, :id
 ]
 
-csv << terms.map(&:to_s).map { |term|
+headers = terms.map(&:to_s).map { |term|
   I18n.translate(:simple_form)[:labels][:generic_file][term] || term.titleize
 }
 
-ActiveFedora::SolrService.query('has_model_ssim:GenericFile', { rows: 99999 }).each.with_index do |gf, idx|
-  gfmod = gf.each.map {|k, v| [k.gsub(/_[a-z]+\z/, ''), v] }.to_h
-  csv << terms.map {|term|
-    t = term.to_s
-    (gfmod[t].is_a?(Array) ? gfmod[t].reject(&:blank?).compact.join(' ; ') : gfmod[t]).to_s
-  }
-  puts "Index: #{idx}" if idx % 100 == 0
+headers << "Events"
+
+csv << headers
+
+def add_metadata_to_csv(model)
+  ActiveFedora::SolrService.query("has_model_ssim:#{model}", { rows: 99999 }).each.with_index do |gf, idx|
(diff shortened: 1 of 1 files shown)


Meowcenary (approved):
From a code perspective seems simple enough, but thanks for putting in the work to figure out how to pull the data from Solr. I'm sure that was a pain, Solr confuses me.

lib/batch/subject.csv.rb
```diff
@@ -1,26 +1,32 @@
...
+      (gfmod[t].is_a?(Array) ? gfmod[t].reject(&:blank?).compact.join(' ; ') : gfmod[t]).to_s
+    }
+    events = model.constantize.find(gfmod["id"]).events
+    event = events.first.blank? ? "no event" : events.first[:action]
+    values << event
+    csv << values
+    puts "Index: #{idx}" if idx % 100 == 0
+  end
 end
+
```
fenekku, line 28: I like the function extraction.

minor: Just some whitespace / syntax changes would make it clearer.

```ruby
def add_metadata_to_csv(model)
  query = ActiveFedora::SolrService.query(
    "has_model_ssim:#{model}", { rows: 99999 }
  )
  query.each.with_index do |gf, idx|
    gfmod = gf.each.map {|k, v| [k.gsub(/_[a-z]+\z/, ''), v] }.to_h

    values = terms.map {|term|
      t = term.to_s
      if gfmod[t].is_a?(Array)
        gfmod[t].reject(&:blank?).compact.join(' ; ')
      else
        gfmod[t].to_s
    }

    events = model.constantize.find(gfmod["id"]).events
    event = events.first.blank? ? "no event" : events.first[:action]
    values << event

    csv << values
    puts "Index: #{idx}" if idx % 100 == 0
  end
end
```

lib/batch/subject.csv.rb
```diff
@@ -1,26 +1,32 @@
...
-  puts "Index: #{idx}" if idx % 100 == 0
-end
-ActiveFedora::SolrService.query('has_model_ssim:Page', { rows: 99999 }).each.with_index do |gf, idx|
-  gfmod = gf.each.map {|k, v| [k.gsub(/_[a-z]+\z/, ''), v] }.to_h
-  csv << terms.map {|term|
-    t = term.to_s
-    (gfmod[t].is_a?(Array) ? gfmod[t].reject(&:blank?).compact.join(' ; ') : gfmod[t]).to_s
-  }
-  puts "Index: #{idx}" if idx % 100 == 0
+headers << "Events"
```
fenekku, line 10: minor: what would happen if "Events" is placed in the `terms` array? Could we save ourselves from these dangling statements?

fenekku (approved):

lib/batch/subject.csv.rb
sharpattack: It's not in solr, so it would break at `gfmod["events"]`. But using your if/else refactor, I can add it to `terms` array, and remove the dangling statements.

lib/batch/subject.csv.rb
sharpattack: Added these suggested changes and incorporated your other comment of removing some dangling statements.

fenekku (approved):
Very nice :heart:

carsonicator (approved):
Thanks @sharpattack

PR #1061, opened 2022-08-12
[#] Null Identifiers For Creator/Contributor

Meowcenary:
When a creator or contributor's orcid identifier is *not* present it
should be omitted. This was accidentally left out due to merge conflicts
with the previous release.

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -215,8 +215,8 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   end
 
   def build_creator_contributor_json(creator)
-    json = @@person_or_org_data[creator]
-    return json if json
+    creatibutor_json = @@person_or_org_data[creator]
+    return creatibutor_json if creatibutor_json
 
     family_name, given_name = creator.split(',', 2)
     given_name = given_name.to_s.lstrip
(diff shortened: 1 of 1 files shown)


Meowcenary:
I was checking the progress on the export and noticed that frustratingly the identifiers field for creatibutors was still showing up even when it was empty. This was almost certainly because of mixup while merging for the release. I decided it was not worth opening a ticket for this since it was such a small and fast change.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -246,14 +244,17 @@ def build_creator_contributor_json(creator)
           "given_name": given_name,
           "family_name": family_name,
           "type": "personal",
-          "identifiers": identifiers
         }
       }
+
+      if dh_user.orcid.present?
+        identifiers = [{"scheme": "orcid", "identifier": dh_user.orcid.split('/').pop}]
+        json.merge!({"identifiers": identifiers})
```
fenekku, line 252: `"identifiers": identifiers` should be merged in `"person_or_org"` value (not top-level).

app/models/invenio_rdm_record_converter.rb
```diff
@@ -246,14 +244,17 @@ def build_creator_contributor_json(creator)
...
+      end
     # Personal record without user in database
     elsif creator.include?(",")
       json = {
-        "person_or_org":
-        {
+        "person_or_org": {
           "given_name": given_name,
           "family_name": family_name,
           "type": "personal"
```
fenekku, line 260: What is the practical difference between this branch and the one above? They seem to result in same output. If nothing else was lost in the merge, then this can be simplified further:

```ruby
if organization?(creator)
...
else
  dh_user = User.find_by(formal_name: creator) || User.find_by(display_name: display_name)
  if dh_user.present?
    ...
    if dh_user.orcid.present?
    ...
    end
  else
    json = {
        "person_or_org": {
          "name": creator,
          "type": "organizational"
        }
      }
end
```

This should also reduce the DB/Solr  calls.

fenekku (commented):
It's always in merge resolution that other issues sneak in :police_officer: .

app/models/invenio_rdm_record_converter.rb
Meowcenary: I think that previously there was significant differences in what we wanted from a creatibutor in the database versus a creatibutor that was not in the database, but yeah with this change we can just remove the branch

app/models/invenio_rdm_record_converter.rb
```diff
@@ -215,8 +215,8 @@ def resource_type(digitalhub_subtype)
   end
 
   def build_creator_contributor_json(creator)
-    json = @@person_or_org_data[creator]
-    return json if json
+    creatibutor_json = @@person_or_org_data[creator]
```
Meowcenary, line 218: Just to make it more clear exactly what json is being referenced without the need to think about scope as much.

fenekku (approved):
Thanks.

(we don't save DB/Solr calls here but the conditional branches are perhaps clearer)

PR #1063, opened 2022-08-15
[#1037] Fix Missing Pediatric Briefs Community Assignment Redux

### LINKED ISSUE #1037: [Export] Fix missing pediatric briefs community assignment
The 2022-04-26 export contains ~2000 records that should be marked as belonging to the `pediatric-neurology-briefs` community, but aren't currently. It looks like records from Pedriatic  Neurology briefs before Volume 15 are not assigned to that community in the export.

Example:

generic_file_pnb-10-9-70-b.json -- https://digitalhub.northwestern.edu/files/pnb-10-9-70-b 
has `"prism_community": ""` in the export but should have `"prism_community": "pediatric-neurology-briefs::Volume 10, Issue 09"`
(e.g. generic_file_pnb-27-2-4.json -- https://digitalhub.northwestern.edu/files/pnb-27-2-4 is good on the other hand with `"prism_community": "pediatric-neurology-briefs::Volume 27, Issue 02"`)

Meowcenary:
The previous iteration expected all PNB collections to follow the same
format, but some PNB collections do not have an issue number which was
causing errors when attempting to convert to an integer. closes #1037

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,11 +76,12 @@ digital-repository/$ cap <stage> deploy
 ```
 
 ##### Digital Hub Export
-* Export repo to JSON formatted for InvenioRDM
+* To export Digital Hub data to JSON format use the below commands from the root project directory. It's critical to use
+		*both* `nohup` AND the trailing ampersand for the first command and similarly important to run the second command
+		with the correct job number. If you don't do all of this then the export will cut out when the terminal that started
+		it in the first place is closed for any reason.
+
 ```
-./bin/rake repo_export
-```
-* Same as above, but with extra debugging information
-```
-./bin/rake verbose debug repo_export
+nohup ./bin/rails r lib/scripts/repo_export_in_batches.rb > "$(date +'%Y-%m-%d-%H%M%S')_repo_export_in_batches.log" &
+disown <job number assigned in previous command>
 ```
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -518,10 +518,17 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
           return mapping_entry
         # part of pediatric neurology brief collection in format pnb-volume#-issue#
         elsif collection_path_entry[:id]&.starts_with?("pnb-")
-          volume_number, issue_number = collection_path_entry[:id].split("-").last(2)
+          volume_number, issue_number = collection_path_entry[:id].split(/[^\d]/).reject(&:blank?)
+
+          if issue_number.present?
+            collection_id = "Volume #{"%02d" % volume_number}, Issue #{"%02d" % issue_number}"
+          else
+            collection_id = "Volume #{"%02d" % volume_number}"
+          end
+
           return {
             "community_id": "pediatric-neurology-briefs",
-            "collection_id": "Volume #{"%02d" % volume_number}, Issue #{"%02d" % issue_number}"
+            "collection_id": collection_id
           }
         end
       end
--- a/spec/models/invenio_rdm_record_converter_spec.rb
+++ b/spec/models/invenio_rdm_record_converter_spec.rb
@@ -1016,9 +1016,15 @@ RSpec.describe InvenioRdmRecordConverter do
     let(:expected_community_collection_string_center_for_file_id) { "science-in-society-scientific-images-contest::2018 Scientific Images Contest Winners" }
     let(:map_file_to_prism_community_collections_irrc) { described_class.new(generic_file_with_community_collection_match, collection_store.data) }
 
+    let!(:collection_pnb) { make_collection(user, title: "Pediatric Neurology Briefs: Volume 05 Issue 07", id: "pnb-5-7", member_ids: [generic_file_pnb.id]) }
     let(:generic_file_pnb) { make_generic_file_with_content(user, id: "pnb-5-7-20") }
     let(:expected_pnb_community_string) { "pediatric-neurology-briefs::Volume 05, Issue 07" }
-    let(:pnb_collection) { described_class.new(generic_file_pnb, collection_store.data) }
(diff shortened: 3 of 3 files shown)


README.md
```diff
@@ -76,11 +76,12 @@ digital-repository/$ cap <stage> deploy
 ```
 
 ##### Digital Hub Export
-* Export repo to JSON formatted for InvenioRDM
-```
-./bin/rake repo_export
-```
-* Same as above, but with extra debugging information
+* To export Digital Hub data to JSON format use the below commands from the root project directory. It's critical to use
```
Meowcenary, line 79: It turns out that the way I ran the command was not the issue with the previous export failing, but I think this is worth putting down here just so we avoid having to relearn this in the future.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -525,10 +525,17 @@ def find_mapping_entry_in_dh_collections
           return mapping_entry
         # part of pediatric neurology brief collection in format pnb-volume#-issue#
         elsif collection_path_entry[:id]&.starts_with?("pnb-")
-          volume_number, issue_number = collection_path_entry[:id].split("-").last(2)
+          volume_number, issue_number = collection_path_entry[:id].split(/[^\d]/).reject(&:blank?)
+
+          if issue_number.present?
```
Meowcenary, line 530: This is the underlying problem - the issue number is not always present which causes errors on formatting

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -1016,9 +1016,15 @@
     let(:expected_community_collection_string_center_for_file_id) { "science-in-society-scientific-images-contest::2018 Scientific Images Contest Winners" }
     let(:map_file_to_prism_community_collections_irrc) { described_class.new(generic_file_with_community_collection_match, collection_store.data) }
 
+    let!(:collection_pnb) { make_collection(user, title: "Pediatric Neurology Briefs: Volume 05 Issue 07", id: "pnb-5-7", member_ids: [generic_file_pnb.id]) }
```
Meowcenary, line 1019: A bit of rearrangement and renaming to hopefully make the testing pattern more clear.

Here you can also see an example of what I mean by a collection id without an issue number in case that was unclear

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -1034,14 +1040,22 @@
...
-        make_collection(user, title: "Pediatric Neurology Briefs: Volume 05, Issue 07", id: "pnb-5-7", member_ids: [generic_file_pnb.id])
-
         collection_store.build_collection_store_data
         collection_store.build_paths_for_collection_store
       end
 
-      it "returns the correct community string" do
-        expect(pnb_collection.send(:dh_collection_to_prism_community_collection)).to eq(expected_pnb_community_string)
+      describe "it returns the correct community string" do
+        context "for a pnb generic file" do
```
Meowcenary, line 1048: Oh , this needs to change.

Meowcenary: I misunderstood some stuff in the specs and put down incorrect information in the context strings, but this is corrected now. The actual specs were fine, just the wrong objects being referred to in the context.

README.md
```diff
@@ -76,11 +76,12 @@ digital-repository/$ cap <stage> deploy
...
+		*both* `nohup` AND the trailing ampersand for the first command and similarly important to run the second command
+		with the correct job number. If you don't do all of this then the export will cut out when the terminal that started
+		it in the first place is closed for any reason.
+
 ```
-./bin/rake verbose debug repo_export
+nohup ./bin/rails r lib/scripts/repo_export_in_batches.rb > "$(date +'%Y-%m-%d-%H%M%S')_repo_export_in_batches.log" &
+disown <job number assigned in previous command>
```
sharpattack, line 86: Does the user need to set the job number here? Or is it just `disown`? 

I prefer to use `screen` for these kind of tasks.

Meowcenary: Just using disown will default to whatever the last started job was so no, it's not necessary. I think it's worth keeping because there's a good chance these will end up getting run out of order at some point OR something else (who knows) that causes an issue. I figure if you are very specific with it less can go wrong.

Meowcenary: And, yes, `screen` will get the job done too from what I can tell, but I've just never really used it so I prefer this way since I know what to expect. If you want to add a section on running it with screen I think that would be good.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -1032,16 +1038,24 @@
...
-        expect(pnb_collection.send(:dh_collection_to_prism_community_collection)).to eq(expected_pnb_community_string)
+      describe "it returns the correct community string" do
+        context "when issue number in collection id" do
+          it "returns the correct community string for a pnb generic file" do
+            expect(generic_file_converter_pnb.send(:dh_collection_to_prism_community_collection)).to eq(expected_pnb_community_string)
+          end
+        end
+
+        context "when issue number not in collection id" do
+          it "returns the correct community string for a" do
```
fenekku, line 1055: minor: 

- "for a"...?

- Also `describe "it returns the correct community string"` and then `it "returns the correct community string..."` is repeated. Is it possible to remove a layer of context/describe?

fenekku (approved):

spec/models/invenio_rdm_record_converter_spec.rb
Meowcenary: Good catch. I've changed the strings to a be more descriptive. 

I think we could remove the string from the spec itself, so `it "test description do` could be replaced with `it do` , but I'm going to leave the strings in for now since they can be helpful for identifying failures 

```
    context "for a pnb generic file" do
        context "that belongs to a collection with an issue number in the id" do
            it "returns the correct community string" do
        context "that belongs to a collection without an issue number in the collection id" do
            it "returns the correct community string" do
```

Issue #1023, opened 2022-06-14
[Export] Original identifier(s)


fenekku:
[Edited 2022-06-15]

DH records with "parsable" original identifiers should export those to Prism as [alternate identifiers](https://inveniordm.docs.cern.ch/reference/metadata/#alternate-identifiers-0-n).

Proposed approach

| DH field                                                          | Prism				 |
| ----------------------------------------------------------------- | ------------------ |
| `original_identifier` or find actual field mapping to original identifier | `identifiers` item |

For DH (https://digitalhub.northwestern.edu/files/bz60cw30w)

Original Identifier
    doi:46589
    (PMID) 11549777

Prism export:

```json
{
  "identifiers": [{
    "identifier": "11549777",
    "scheme": "pmid"
  }]
}
```

The original identifier DOI in that example is not a valid DOI (a valid doi is `<prefix>/<suffix>` with some rules), so it doesn't make sense to export. For DOIs, just checking against the shape in the export should be enough. On the import, we can use an already developed more sophisticated checker if need be.  

**Questions**

- What to do with original identifiers that don't map to a controlled scheme e.g. PNB-19-01 ?
- What to do with original identifiers created for GV Black e.g., m19091029 ? @gneidhardt  will see Katie [EDIT from [comment](https://github.com/galterlibrary/digital-repository/issues/1023#issuecomment-1156490325) ignore those as they are for the GV Black collections that should not be migrated]

gneidhardt:
Unfortunately, this field is not included in the current file export from DigitalHub. @sharpattack, is this a field we can include on the next export? I can then see what all the examples are for the whole dataset. 

Looking at just the 50 files we sampled, these largely seem to be:

- PMID (pubmed ids) for some papers (example: (PMID) 34109874)
- article ids for the Pediatric Neurology Briefs (example: PNB-29-92), which will be numerous as there are 1000s of PNB articles
- original identifiers created for the GV Black collection (example: m19091029) - likely we'll need to consult with Katie on whether or not these need to stay in the record.

Meowcenary:
@gneidhardt - I'm planning on getting to this ticket in the near future. Do you still want/need the export of the entires that don't conform to a standard schema? If so I'll enter a separate ticket for that.

Meowcenary:
Pointing this relatively high since free text is usually difficult to work with. Probably going to involve some regex or something similar.

Meowcenary:
[...]

PR #1064, opened 2022-08-15
[#1023] Original Identifiers

Meowcenary:
Export DH identifiers that are part of a recognized scheme

Previously this was blocked while we waited for answers about files related to the 12th General Hospital, but it turns out that the files are not needed.

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -183,7 +183,7 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
         "contributors": contributors(@generic_file.contributor),
         "dates": format_dates(@generic_file.date_created),
         "languages": @generic_file.language.map{ |lang| LANGUAGES[lang.downcase] ? {"id": LANGUAGES[lang.downcase]} : nil }.compact,
-        "identifiers": ark_identifiers(@generic_file.ark),
+        "identifiers": original_identifiers(@generic_file.identifier) + ark_identifiers(@generic_file.ark),
         "related_identifiers": related_identifiers(@generic_file.related_url),
         "sizes": Array.new.tap{ |size_json| size_json << "#{@generic_file.page_count} pages" if !@generic_file.page_count.blank? },
         "formats": [@generic_file.mime_type],
@@ -546,5 +546,33 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
     else
       community_id
     end
+  end
+
+  def original_identifiers(identifiers)
+    identifiers = identifiers.map do |identifier|
+      if identifier.include?("PMID")
+        id = identifier.gsub(/\(PMID\)/, "").strip
+        scheme = "pmid"
+      elsif identifier.include?("DOI")
+        # it's significantly less difficult to just do two removals than find a catch all here
+        id = identifier.gsub(/DOI/, "").gsub(/[():]/, "").strip
+        scheme = "doi"
+      elsif identifier.include?("ISBN")
+        id = identifier.gsub(/\(ISBN.*\)/, "").strip
+        scheme = "isbn"
+      elsif identifier.include?("PNB")
+        id = identifier
+        scheme = "other"
+      else
+        next
+      end
+
+      {
+        "identifier": id,
+        "scheme": scheme
+      }
+    end
+
+    identifiers.compact
   end
 end
--- a/spec/models/invenio_rdm_record_converter_spec.rb
+++ b/spec/models/invenio_rdm_record_converter_spec.rb
@@ -34,7 +34,7 @@ RSpec.describe InvenioRdmRecordConverter do
       mesh: [mesh_term],
       lcsh: [lcsh_term],
       subject_geographic: ["Boston (Mass.)", "Chicago (Ill.)"],
-      based_near: ['Boston, Massachusetts, United States', 'East Peoria, Illinois, United States'],
+      based_near: ['East Peoria, Illinois, United States', 'Boston, Massachusetts, United States'],
       description: ["This is a generic file for specs only", "This is an additional description to help test"],
       date_created: ["2021-1-1"],
       mime_type: 'application/pdf',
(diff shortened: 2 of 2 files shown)


app/models/invenio_rdm_record_converter.rb
```diff
@@ -547,4 +547,29 @@ def format_prism_community_collection_string(mapping_entry)
...
+  def original_identifiers(identifiers)
+    identifiers.map do |identifier|
+      if identifier.include?("PMID")
+        id = identifier.gsub(/\(PMID\)/, "").strip
+        scheme = "pmid"
+      elsif identifier.include?("DOI")
+        id = identifier.gsub(/\("DOI"\)/, "").strip
+        scheme = "doi"
+      elsif identifier.include?("ISBN")
+        id = identifier.gsub(/\(ISBN}.*\)/, "").strip
```
fenekku, line 560: The ISBN identifiers have a `}` in them ?

app/models/invenio_rdm_record_converter.rb
```diff
@@ -547,4 +547,29 @@ def format_prism_community_collection_string(mapping_entry)
...
+        scheme = "other"
+      else
+        next
+      end
+
+      {
+        "identifier": id,
+        "scheme": scheme
+      }
+    end
```
fenekku, line 573: Put the `compact` call from L186 in the method itself to remove dangling operation.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -34,7 +34,7 @@
       mesh: [mesh_term],
       lcsh: [lcsh_term],
       subject_geographic: ["Boston (Mass.)", "Chicago (Ill.)"],
-      based_near: ['Boston, Massachusetts, United States', 'East Peoria, Illinois, United States'],
+      based_near: ['East Peoria, Illinois, United States', 'Boston, Massachusetts, United States'],
```
fenekku, line 37: What changed here?

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -45,7 +45,9 @@
       visibility: InvenioRdmRecordConverter::OPEN_ACCESS,
       related_url: ["https://doi.org/10.5438/55e5-t5c0"],
       acknowledgments: ["this is an acknowledgement"],
-      abstract: ["this is an abstract"]
+      abstract: ["this is an abstract"],
+      identifier: ["(DOI):46589", "(PMID) 11549777", "cruisemicrobe", "615.778-2", "600-15", "m19090127", "PNB-23-38", \
+                     "HSL.2016.15.0007", "(ISBN 10) 3956501241", "PNB-14-96-a", "PNB-15-10"]
```
fenekku, line 50: minor: let's place them one per line for readability (limiting line length between 80 and 90 characters is typically a good size)

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -164,9 +166,29 @@
...
+              "scheme": "ark"
+            },
+            {
+              "identifier": "PNB-14-96-a",
+              "scheme": "other"
+            },
+            {
+              "identifier": "PNB-15-10",
+              "scheme": "other"
+            }],
```
sharpattack, line 191: Is `PNB-23-38` missing?

app/models/invenio_rdm_record_converter.rb
Meowcenary: Nice catch. For a second I thought it might be legit, but I think it would have been added to the specs if it were necessary.

spec/models/invenio_rdm_record_converter_spec.rb
Meowcenary: Agree on one per a line, but I generally prefer ~120 characters. 80 is just too tight IMO

spec/models/invenio_rdm_record_converter_spec.rb
Meowcenary: Honestly, I have no clue what happened, but for some reason the test suite expected them in reversed order on this branch. None of the objects used for testing are changed here so I'm not really sure what to think.

spec/models/invenio_rdm_record_converter_spec.rb
Meowcenary: Oh, good catch. That test case isn't necessary anymore since it follows the same pattern as "PNB-15-10" of "PNB-integer-integer".

app/models/invenio_rdm_record_converter.rb
Meowcenary: good call

Meowcenary:
All the comments have been addressed, but there is a failing spec that I need to figure out still.

spec/models/invenio_rdm_record_converter_spec.rb
fenekku: Ha! the python PEP8 line size limit is 80 (120 is a 50% increase!), black -a modern formatting tool- puts it at 88 and GitHub seems to display 94-95 char lines. 

Again, I don't mind so much, as long as the result is clean and readable.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -547,4 +547,32 @@ def format_prism_community_collection_string(mapping_entry)
...
   end
+
+  def original_identifiers(identifiers)
+    identifiers = identifiers.map do |identifier|
+      if identifier.include?("PMID")
+        id = identifier.gsub(/\(PMID\)/, "").strip
+        scheme = "pmid"
+      elsif identifier.include?("DOI")
+        # it's significantly less difficult to just do two removals than find a catch all here
+        id = identifier.gsub(/DOI/, "").gsub(/[():]/, "").strip
```
fenekku, line 558: :crossed_fingers: there is no DOI called "10.1234/IDOIT_CAUSE_I_CAN" ;)

fenekku (approved):

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -1053,7 +1085,6 @@
 
         context "that belongs to a collection without an issue number in the collection id" do
           it "returns the correct community string" do
-            expect(generic_file_converter_pnb_short_id.send(:dh_collection_to_prism_community_collection)).to eq(expected_pnb_short_id_community_string)
```
Meowcenary, line 1056: This should not have been removed...

app/models/invenio_rdm_record_converter.rb
Meowcenary: That would be miserable lol. I need to take a class on common techniques for dealing with string parsing and identification. I'm sure these kinds of things come up for other programmers too.

PR #1044, opened 2022-07-11
Release 1.27.0 Combine Descriptions, Update Header Lookup, Map Geographic Subjects And Location 

Meowcenary:
Combine descriptions, update header lookup to read jsonl formatted files line by line rather than loading into memory all at once, update the export to map geographic subjects and presentation location

PR #1066, opened 2022-08-16
Release 1.29.1-mid_august_2022_export_updates

Meowcenary:
Small fixes for problems with the export most important of which is a fixing a breaking bug with identifying Pediatric Neurology Briefs

Going to double check this since there were some frustrating merge conflicts, but the specs are green on my local instance and from what I can tell all the commits have been applied.

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -215,8 +215,8 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   end
 
   def build_creator_contributor_json(creator)
-    json = @@person_or_org_data[creator]
-    return json if json
+    creatibutor_json = @@person_or_org_data[creator]
+    return creatibutor_json if creatibutor_json
 
     family_name, given_name = creator.split(',', 2)
     given_name = given_name.to_s.lstrip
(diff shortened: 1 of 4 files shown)


Meowcenary:
Not sure what's going on with all the merge conflicts, but I rebased release off of production which seems to have sorted out the issues.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -227,54 +227,47 @@ def build_creator_contributor_json(creator)
...
           "type": "personal"
         }
       }
+
+      if dh_user.orcid.present?
+        identifiers = [{"scheme": "orcid", "identifier": dh_user.orcid.split('/').pop}]
+        creatibutor_json["person_or_org"].merge!({"identifiers": identifiers})
+      end
+    # Personal record without user in database
     # Unknown / Not Identified creator
```
Meowcenary, line 255: I'll remove this comment quickly

Meowcenary (commented):
Did a quick review for my own sake and it seems good to me.

app/models/invenio_rdm_record_converter.rb
Meowcenary: Eh, I'm just going to leave it. I'd have to add the commit outside of release or we'd end up with more annoying merge conflicts in the future.

Meowcenary:
Testing this on staging now. The actual app is working fine, now Ill check the export quickly.

We might need to close this and open another release branch now that I have a handle on the rebasing issue. Just waiting for some feedback.

fenekku (approved):
Looks ok :shipit:

sharpattack:
Seems off to me. This is rebased off of production branch. Starting with this PR https://github.com/galterlibrary/digital-repository/pull/1044 , production/release branch did not have commits from master.

I think we may always have merge conflicts if we want to continue forward. We should address this now. Sorry, but if it's not urgent, please give me a day to think about how to fix this so we can have smoother releases in the future. If it can't wait, then let it be known that at least the export may not be accurate if it's run on this code base.

carsonicator (approved):
Thanks @Meowcenary

Meowcenary:
I think the reason for the conflicts was due to commits that were previously added being changed by new branches. I went through rebase work myself with the following:

```
git checkout release
git pull --rebase origin production # origin can be omitted if you have the changes locally
# this raises conflicts which are all resolved by repeating the below command 
git rebase --skip
```

The reason `git rebase --skip` works is because the work in the old commits that have conflicts is also in later commits. All the code I'm expecting to be on this branch is here.

Meowcenary:
Here we goooooooo

PR #1069, opened 2022-08-17
[#1068] 2022-08-17 Repo Export Bug Fixes

### LINKED ISSUE #1068: 2022-08-17 Repo Export Bug Fixes
- InveinoRdmRecordConverter#build_creator_contributor_json has a logic issue where it expects have a user object to check for an orcid id, but the value is nil. Add safe navigation operator to this
- Seemingly related to the above, there's an error that appears a few times where `.merge!({...})` is being called on a nil value. The most obvious place to check is build_creator_contributor_json, but this issue might be a result of a nil value being returned from here to a different function
- ... more to come I'm sure

Meowcenary:
Previous refactors introduced a bug where a User object is expected, but
nil values were being referenced. Also in previous refactors creatibutor
json was updated to have identifiers under the key "person_or_org". Ruby
hashes can be constructed with strings as key, but must be looked up
with symbols. closes #1068

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -241,15 +241,23 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
     elsif dh_user.present? || creator.include?(",")
       creatibutor_json = {
         "person_or_org": {
+          "given_name": "given_name",
+          "family_name": "family_name",
+          "type": "personal"
+        }
+      }
+
+      creatibutor_json = {
+        "person_or_org": {
           "given_name": given_name,
           "family_name": family_name,
           "type": "personal"
         }
       }
 
-      if dh_user.orcid.present?
+      if dh_user&.orcid.present?
         identifiers = [{"scheme": "orcid", "identifier": dh_user.orcid.split('/').pop}]
-        creatibutor_json["person_or_org"].merge!({"identifiers": identifiers})
+        creatibutor_json[:person_or_org].merge!({"identifiers": identifiers})
       end
     # Personal record without user in database
     # Unknown / Not Identified creator
--- a/app/models/concerns/galtersufia/generic_file/known_organizations.rb
+++ b/app/models/concerns/galtersufia/generic_file/known_organizations.rb
@@ -7,6 +7,7 @@ module Galtersufia
         "Galter Health Sciences Library",
         "12th General Hospital",
         "Northwestern University, Medical School",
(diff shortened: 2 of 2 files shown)


app/models/invenio_rdm_record_converter.rb
```diff
@@ -247,9 +255,9 @@ def build_creator_contributor_json(creator)
         }
       }
 
-      if dh_user.orcid.present?
+      if dh_user&.orcid.present?
```
fenekku, line 258: My bad with that `|| creator.include?(",")` refactor.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -247,9 +255,9 @@ def build_creator_contributor_json(creator)
...
         identifiers = [{"scheme": "orcid", "identifier": dh_user.orcid.split('/').pop}]
-        creatibutor_json["person_or_org"].merge!({"identifiers": identifiers})
+        creatibutor_json[:person_or_org].merge!({"identifiers": identifiers})
```
fenekku, line 260: yeah pretty awful

fenekku (approved):
Thanks for fixing those right away @Meowcenary

app/models/invenio_rdm_record_converter.rb
```diff
@@ -239,6 +239,14 @@ def build_creator_contributor_json(creator)
       }
     # Personal record with user in database OR is a person's formal name
     elsif dh_user.present? || creator.include?(",")
+      creatibutor_json = {
+        "person_or_org": {
+          "given_name": "given_name",
+          "family_name": "family_name",
+          "type": "personal"
+        }
+      }
```
sharpattack, line 248: Is this supposed to be in here?

app/models/invenio_rdm_record_converter.rb
Meowcenary: Nah, that's my bad for not noticing lol. I wrote this myself after all.

app/models/invenio_rdm_record_converter.rb
Meowcenary: omg no :(

Meowcenary: ahhhh thanks @sharpattack  - this would have been disastrous

sharpattack: I see this was merged. Did this change end up being needed?

PR #1072, opened 2022-08-22
1022 Export Custom Fields Number In Sequence And Original Citation

### LINKED ISSUE #1022: [Export] Export custom fields `number_in_sequence` and `original_citation` as additional_descriptions
Realistically, custom fields won't be ready for initial launch. In a migration meeting, we decided to export `number_in_sequence` and `original_citation` via additional descriptions.

Here is the approach (comment below to discuss it!):

Export the corresponding DH fields as Prism `additional_descriptions` items.

| DH field                                                          | Prism				 |
| ----------------------------------------------------------------- | ------------------ |
| `page_number` or `page_number_actual` (probably not `page_count`) | number_in_sequence |
| `bibliographic_citation` 											| original_citation  |

DH fields are to be double-checked if they are the right ones.

The added additional_description item will look like:

```json
    {
      "description": "number_in_sequence: <value>",
      "type": {
        "id": "other"
      }
    },
```
AND/OR

```json
    {
      "description": "original_citation: <value>",
      "type": {
        "id": "other"
      }
    }
```

For example, for a DH record with `page_number` of value `3` and `bibliographic_citation` of value `Custom original citation` and no other `additional_descriptions`, the export for `additional_descriptions` will look like:

```json
"additional_descriptions": [
  {
    "description": "number_in_sequence: 3",
    "type": {
      "id": "other"
    }
  },
  {
    "description": "original_citation: Custom original citation",
    "type": {
      "id": "other"
    }
  }
]
```

Meowcenary:
Export digitalhub page_number and bibliographic_citation as additional descriptions. closes #1022

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -10,7 +10,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   SUBJECT_FIELDS = [:tag, :mesh, :lcsh, :subject_name, :subject_geographic]
   CIRCA_VALUES = ["ca.", "ca", "circa", "CA.", "CA", "CIRCA"]
   UNDATED = ["undated", "UNDATED"]
-  REFERENCE_FIELDS = ["bibliographic_citation", "part_of"]
   ABBR_MONTHNAMES = Date::ABBR_MONTHNAMES.map{ |abbr_monthname| abbr_monthname.downcase if abbr_monthname.present? }
   MONTHNAMES = Date::MONTHNAMES.map{ |monthname| monthname.downcase if monthname.present? }
   SEASONS = ["spring", "summer", "fall", "winter"]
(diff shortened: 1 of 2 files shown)


Meowcenary:
Oh, let me squash this first.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -10,7 +10,6 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   SUBJECT_FIELDS = [:tag, :mesh, :lcsh, :subject_name, :subject_geographic]
   CIRCA_VALUES = ["ca.", "ca", "circa", "CA.", "CA", "CIRCA"]
   UNDATED = ["undated", "UNDATED"]
-  REFERENCE_FIELDS = ["bibliographic_citation", "part_of"]
```
Meowcenary, line 13: I think I had left this in thinking that we might need to use it again, but it's best to just remove it entirely.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -176,7 +175,9 @@ def invenio_metadata
         "additional_descriptions":
           format_additional("description", "acknowledgements", @generic_file.acknowledgments) +
           format_additional("description", "abstract", @generic_file.abstract) +
-          format_additional("description", "other", @generic_file.based_near, "presentation_location: "),
+          format_additional("description", "other", @generic_file.based_near, "presentation_location: ") +
+          [{"description":  "number_in_sequence: #{@generic_file.page_number.to_s.force_encoding("UTF-8")}", "type": {"id": "other"}}] +
```
Meowcenary, line 179: Weirdly, `page_number` is a string and not an array. It seems that `page_number_actual` is stored as an integer and used for Solr queries and sorting. `page_number_actual`, from what I saw in the specs, will only be present if `page_number`  does not have any characters in it. E.g something like "29a" will result in a blank `page_number_actual` , but I think this is just to prevent errors with Solr.

fenekku (approved):
LGTM

PR #1075, opened 2022-08-22
[#] Resyncing Commits Between Master And Prod

Meowcenary:
The latest release will resync master and production so there are a lot of commits going in. This removes some code that should not be left in, but has no effect.

After merging this we can open the release PR which will have the many commits needed to resync.


fenekku (approved):
That works.

PR #1080, opened 2022-08-24
[#1079] Remove Empty Number In Sequence

### LINKED ISSUE #1079: [export] Remove empty number_in_sequence
The vast majority of exported number_in_sequence is empty i.e. `"description": "number_in_sequence: "` . We should not generate a description entry for those.

Meowcenary:
Refactor to use format_additional function which handles empty values
closes #1079

`format_additional` checks each value to see if it's blank. If page_number is blank or nil then `[@generic_file.page_number.to_s.force_encoding("UTF-8")]` will return `[""]` which will then be skipped in `format_additonal`


fenekku (approved):
Thanks

PR #1082, opened 2022-08-26
[#1078] Normalize ISBN Identifiers To One

### LINKED ISSUE #1078: Normalize ISBN identifiers to 1 (ISBN13 format)
From 2022-08-23 export

Only distinct metadata.identifiers schemes are allowed in InvenioRDM. File generic_file_4f271d03-e6e4-4bd5-b72f-2b1ec679483c.json has:

```
      "identifiers": [
        {
          "identifier": "3956501241",
          "scheme": "isbn"
        },
        {
          "identifier": "9783956501241",
          "scheme": "isbn"
        },
        {
          "identifier": "ark:/c8131/g32g7x",
          "scheme": "ark"
        }
      ],
```
i.e. 2 ISBNs . 9783956501241 is the ISBN13 version of 3956501241

Just export the ISBN13 in situations like these where there are 2 equivalent ISBNs. Maybe Ruby has a library for ISBN manipulation. InvenioRDM uses isbnlib -relevant code: https://github.com/xlcnd/isbnlib/blob/1886ae31b88caaae1557b9159a92fec0af99af54/isbnlib/_core.py#L111 .

Meowcenary:
When parsing identifiers if more than one ISBN identifiers are found
remove the ten digit isbn in favor of the thirteen digit isbn.
closes #1078


app/models/invenio_rdm_record_converter.rb
```diff
@@ -574,6 +577,21 @@ def original_identifiers(identifiers)
       }
     end
 
+    # if there are multiple isbn find the isbn with length 10 and remove it
+    if multiple_isbn
+      identifiers = normalize_isbn(identifiers)
```
Meowcenary, line 582: Alternatively we could do this

```
identifiers = identifiers.compact.map{ |id_obj| id_obj[:scheme] == "isbn" && id_obj[:identifier].to_s.length == 10 ? nil : id_obj }
```

Meowcenary (commented):
I think we could replace the function with a ternary operator, but not sure if it's really an improvement. Otherwise simple PR

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -207,18 +208,19 @@
               "identifier": "PNB-14-96-a",
               "scheme": "other"
             },
-            {
-              "identifier": "3956501241",
-              "scheme": "isbn"
-            },
             {
               "identifier": "PNB-15-10",
               "scheme": "other"
```
fenekku, line 213: Hmmm, is it possible to get a repeat "other" in the export? I didn't encounter a record with a repeat "other" identifier scheme in the latest export, so we are good for now. But if there is a repeat, we would probably be in the same position... If the "other" scheme is never repeated (and it seems like it so far), we don't have to do anything about it.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -574,6 +577,21 @@ def original_identifiers(identifiers)
...
   end
+
+  def normalize_isbn(identifiers)
+    identifiers.map do |id_obj|
+      if id_obj.present? && id_obj[:scheme] == "isbn" && id_obj[:identifier].to_s.length == 10
+        nil
+      else
+        id_obj
+      end
+    end
```
fenekku, line 595: The issue here is that if the record only has 1 ISBN10, it will be removed. `multiple_isbn` is set to true when there is a single isbn instead of when there are multiple. It's fair to assume there is at most only 1 ISBN of each type per record though.

fenekku (commented):
A question and something to adjust, thanks.

spec/models/invenio_rdm_record_converter_spec.rb
Meowcenary: The only original identifiers that are mapping to other right now are those that have PNB in them. I think that only a single PNB identifier is applied to each file since they seem to follow a pattern which is meant to identify which brief it's from and what page it is (or something along those lines hehe)

app/models/invenio_rdm_record_converter.rb
```diff
@@ -561,6 +563,7 @@ def original_identifiers(identifiers)
       elsif identifier.include?("ISBN")
         id = identifier.gsub(/\(ISBN.*\)/, "").strip
         scheme = "isbn"
+        isbn_count += 1
```
Meowcenary, line 566: Increment each time an ISBN is found rather than setting this as true which would lead to the bug @fenekku pointed out below. This seems like a simple enough approach, but I don't love it.

fenekku (approved):
Thanks for the changes.That will do as we are wrapping up the export.

PR #1091, opened 2022-09-19
[#1070] Disable File Uploads

### LINKED ISSUE #1070: Disable Uploads
Switch to Prism, October 23, 2022.

We should add a banner to notify users of some info.

Meowcenary:
Disable file uploads by setting max allowed files to zero, max file size to zero bytes, disabling the buttons that are used for file upload, and removing all links to the file upload page. closes #1070


app/views/generic_files/_browse_everything.html.erb
```diff
@@ -1,30 +1,2 @@
-<%= render partial: 'generic_files/upload/agreement' %>
-<div class="alert alert-success">
-  <%=  t("sufia.upload.cloud_timeout_message", contact_href: link_to(t("sufia.upload.alert.contact_href_text"), sufia.contact_form_index_path)) %>
+<div >
```
Meowcenary, line 1: This and the other empty view are here to override the views that are provided by sufia

app/views/dashboard/_index_partials/_heading_actions.html.erb
```diff
@@ -0,0 +1,14 @@
+<div class="col-xs-12 heading-row">
```
Meowcenary, line 1: This is to override the sufia provided view

app/assets/javascripts/files.js
```diff
@@ -1,8 +1,6 @@
 $(function () {
```
Meowcenary, line 1: Disabling the file upload

fenekku (approved):
Works for me. People should still be able to edit (which is what we want, and in an ideal world, want only for librarians but that's ok), right?

Obviously this should not be merged/deployed until the opportune moment.

Meowcenary:
> Works for me. People should still be able to edit (which is what we want, and in an ideal world, want only for librarians but that's ok), right?
> 
> Obviously this should not be merged/deployed until the opportune moment.

From what I saw the editing would still work fine. Yes, we're just going to have this ready for when the time comes.

carsonicator (approved):
Thanks @Meowcenary . Yes, ideally, we would like to allow edit access only to certain users after these "read-only" changes are made.

mattbaumann1:
@Meowcenary went to look at another pull request and noticed this pull request hanging out. Is a review on this still needed?

fenekku:
This PR is just waiting on the go from the decision team to make the transition to Prism before being merged. When it's merged and deployed, DigitalHub new entries will be disabled.

mattbaumann1 (approved):

Meowcenary:
Going to rebase this today since it has been sitting out here for awhile. I don't think there will be conflicts, but better to just stay up to date.

Meowcenary:
I just rebased this so it should be good to go whenever we want to disable uploading. I'll test it out after I finish changes for #1117

PR #1106, opened 2022-09-22
[#97] Change Josh's Email In Metadata

### LINKED ISSUE #97: Integrate with google traffic tracker

Meowcenary:
Update InvenioRdmRecordConverter#owner_info to take the username as an argument and to check specifically for Josh's email that is then replaced with the proper casing. closes #97

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -84,20 +84,21 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
     data
   end
 
-  def owner_info
-    user = User.find_by(username: @generic_file.depositor)
+  def owner_info(depositor)
+    user = User.find_by(username: depositor)
+    user_email  = user.email == "joshelder@northwestern.edu" ? "JoshElder@northwestern.edu" : user.email
 
     if user
-      { user.username => user.email }
+      {user.username => user_email}
     else
-      { "unknown": "unknown" }
+      {"unknown": "unknown"}
     end
   end
 
   def file_permissions
     permission_data = Hash.new
 
-    permission_data["owner"] = owner_info
+    permission_data["owner"] = owner_info(@generic_file.depositor)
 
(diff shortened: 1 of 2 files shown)


app/models/invenio_rdm_record_converter.rb
```diff
@@ -84,20 +84,21 @@ def extra_data
     data
   end
 
-  def owner_info
-    user = User.find_by(username: @generic_file.depositor)
+  def owner_info(depositor)
```
Meowcenary, line 87: Adding the argument makes it easier to test because you can then re-use an instance of InvenioRdmRecordConverter without running all the setup again.

fenekku: For digitalhub at this point it's fine. I think in general this is probably a sign that `InvenioRdmRecordConverter` is lacking an easy creation or that the tests need rearranging.

fenekku (approved):
Works for me given caveat. I will do the equivalent change in galter-ir-importer communities.json for his email too.

spec/models/invenio_rdm_record_converter_spec.rb
```diff
@@ -1153,4 +1152,12 @@
       end
     end
   end
+
+  describe "owner_info" do
+    let(:josh_elder_user) { FactoryGirl.create(:user, username: "josh_the_elder", formal_name: "Elder, Josh",  display_name: 'Josh Elder', email: 'joshelder@northwestern.edu') }
```
sharpattack, line 1157: 🧙

sharpattack (approved):
I was going to suggest just changing Josh's email from the database, but I think it's configured to downcase on save/update.

sharpattack:
Testing new slack notification. Ignore.

spec/models/invenio_rdm_record_converter_spec.rb
Meowcenary: I really like this emoji haha

PR #1111, opened 2022-09-26
[#1110] Sepetember 2022 Release Bug Fixes

### LINKED ISSUE #1110: September 2022 Release Export Bug Fixes
Error : undefined method merge! for nil:NilClass
Some example files: qb98mf449, 6t053f96k, gt54kn002, ws859f64s

Error - undefined method include? for nil:NilClass
Some example files: p2676v53n, rv042t05q, b8515n36q

The migration is still running so if more errors come up I'll add them here

Meowcenary:
Modify export script to skip pages entirely to clear up the error log and save time. Modify Invenio record converter creatibutor json creation to use with_indifferent_access for adding orcid identifiers. closes #1110


app/models/invenio_rdm_record_converter.rb
```diff
@@ -253,7 +253,7 @@ def build_creator_contributor_json(creator)
 
       if dh_user&.orcid.present?
         identifiers = [{"scheme": "orcid", "identifier": dh_user.orcid.split('/').pop}]
-        creatibutor_json[:person_or_org].merge!({"identifiers": identifiers})
+        creatibutor_json.with_indifferent_access[:person_or_org].merge!({"identifiers": identifiers})
```
Meowcenary, line 256: Hashes have been causing annoying issues due to keys as symbols or strings. This is ugly, but based on my testing in the terminal it will get return what we want and work for symbols or keys hopefully resolving this issue for good.

lib/scripts/repo_export_in_batches.rb
```diff
@@ -35,6 +35,8 @@
   conversion_count = 0
 
   converter[:model_class].find_each do |record_for_export|
+    next if record_for_export.is_a?(Page)
```
Meowcenary, line 38: No more will we waste time exporting pages!

app/models/invenio_rdm_record_converter.rb
fenekku: ugh sorry you have to go through this with Ruby :grimacing:

lib/scripts/repo_export_in_batches.rb
fenekku: Hmmm in the last imports I got there were no `page_<identifier>.json` files. So either we were already doing that filtering somewhere else (perhaps not as efficiently as this though) or this will cull more than we want. Worth taking another look.

mattbaumann1 (approved):

Meowcenary:
Decided to move the check for if the GenericFile is a Page object to the export script itself since the `unexportable?` method on GenericFile needs to know what collections the GenericFile belongs to which occurs during all the other expensive metadata building work.

PR #1112, opened 2022-09-26
[#1088] PDF Page Counts For Some Records Is Off

### LINKED ISSUE #1088: [Export] PDF page count for some records is off
This is an error both in DH and Prism, so possibly a bug with the file. PDF file page count in both DH and Prism is "0 pages", but this PDF has 44 pages.

https://vtfsmghslrepo02.fsm.northwestern.edu/records/9zh4q-y4k87
https://digitalhub.northwestern.edu/files/9cd9e3df-458c-4a2c-9e42-1dcdc95e7adf

Screenshot of DH (it's in admin data):
image

Meowcenary:
Specify page count for records with erroneous pdf page counts. closes #1088

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -189,7 +189,7 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
         "languages": @generic_file.language.map{ |lang| LANGUAGES[lang.downcase] ? {"id": LANGUAGES[lang.downcase]} : nil }.compact,
         "identifiers": original_identifiers(@generic_file.identifier) + ark_identifiers(@generic_file.ark),
         "related_identifiers": related_identifiers(@generic_file.related_url),
-        "sizes": Array.new.tap{ |size_json| size_json << "#{@generic_file.page_count} pages" if !@generic_file.page_count.blank? },
+        "sizes": sizes,
         "formats": [@generic_file.mime_type],
         "version": version(@generic_file.content),
         "rights": rights(@generic_file.rights),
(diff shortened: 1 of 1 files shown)


app/models/invenio_rdm_record_converter.rb
```diff
@@ -611,4 +611,14 @@ def publisher
   def pediatric_neurology_brief?
     @generic_file.depositor == INSTITUTIONAL_PNB_DEPOSITOR || @generic_file.doi.shift.include?(PNB_DOI_PREFIX)
   end
+
+  def sizes
+    if @generic_file.id == "3105875f-61e1-412c-9b1c-c8a33b37ff35"
```
Meowcenary, line 616: This seems like the fastest and simplest solution. If we find more situations like this we can just add more elif statements.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -611,4 +611,14 @@ def publisher
...
+      ["116 pages"]
+    elsif  @generic_file.id == "9cd9e3df-458c-4a2c-9e42-1dcdc95e7adf"
+      ["44 pages"]
+    elsif @generic_file.page_count
+      ["#{@generic_file.page_count} pages"]
```
fenekku, line 621: This is another issue with pages that we should fix: `@generic_file.page_count` is an array so the final string looks like:

`'["0"] pages'`

when it should be:

`'0 pages'`

(e.g., https://vtfsmghslrepo02.fsm.northwestern.edu/records/653g4-8m023)
Although these 0 pages shouldn't really be the norm actually - was `@generic_file.page_count.blank?` discarded for a specific reason?

Popping the page count from `page_count` should probably be enough.

Meowcenary: Oh, I think that was actually on the previous version. I was wondering why that was there.

Meowcenary: Okay, updated!

app/models/invenio_rdm_record_converter.rb
```diff
@@ -611,4 +611,14 @@ def publisher
   def pediatric_neurology_brief?
     @generic_file.depositor == INSTITUTIONAL_PNB_DEPOSITOR || @generic_file.doi.shift.include?(PNB_DOI_PREFIX)
   end
+
+  def sizes
+    if @generic_file.id == "3105875f-61e1-412c-9b1c-c8a33b37ff35"
+      ["116 pages"]
+    elsif  @generic_file.id == "9cd9e3df-458c-4a2c-9e42-1dcdc95e7adf"
+      ["44 pages"]
+    elsif @generic_file.page_count
```
sharpattack, line 620: You need to check if this is blank, right?

Meowcenary: Nope, we want to check if there is anything to format. If it's blank we will just skip it.

sharpattack: Whoops, I meant not blank. Empty arrays are true, right?

Meowcenary: oh right! that was the reason it was using that logic

fenekku (approved):
Updates look good to me. Thanks!

PR #1118, opened 2022-10-07
[#] Update GV Black Collection To Correct ID

Meowcenary:
This is based on what we discussed at today's meeting. The golden egg as @fenekku called it 😄 

Previously the wrong id was being used, but the logic surrounding what should be exported from within is not changed.


sharpattack (approved):
😓 how did I not check this the first time around?

fenekku (approved):
:swan: :egg: :1st_place_medal:

Meowcenary:
> 😓 how did I not check this the first time around?

Don't worry about it. This was a super confusing part of the export.

PR #1119, opened 2022-10-10
Update dh_collections_prism_collection_community.json

kglibrarian:
Added "Chicago Area Medical Archivists Medical History Symposium" as a new collection. Reviewing and adding collections to researchers' collection (not yet finished review).

### DIFF
--- a/dh_collections_prism_collection_community.json
+++ b/dh_collections_prism_collection_community.json
@@ -76,6 +76,10 @@
     "collection_id": "A retrospective look at the predictions and recommendations from the 2009 AMIA Policy Meeting",
     "community_id": "researchers-collections"
   },
+   "e0338411-7829-49ac-8fdc-cd17b7307474": {
+    "collection_id": "AASLD 2018",
+    "community_id": "researchers-collections"
+  },
   "d0798568-47c3-453e-ae39-242d8a96b1dc": {
     "collection_id": "Aberrant cGMP Signaling Persists During Recovery in Mice with Oxygen-Induced Pulmonary Hypertension",
     "community_id": "researchers-collections"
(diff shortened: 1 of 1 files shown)


dh_collections_prism_collection_community.json
```diff
@@ -419,7 +427,8 @@
   "3635f61e-e67e-41f2-b4de-982a9f81dcc8": {
     "community_id": "covid-19-community"
   },
-  "5f1b1739-512f-4015-98bc-22f37f42af7b": {
-    "community_id": "calantone-et-al-2022-commun-biol-source-data"
+
```
fenekku, line 430: minor: let's remove the empty line

fenekku (approved):
@kglibrarian Is this complete or is this still a work-in-progress (with respect to the mapping you know about)? If complete, once the (very) minor comment is addressed this can be merged.
Otherwise let us know what else will be needed to get this merged in.

dh_collections_prism_collection_community.json
```diff
@@ -68,18 +68,26 @@
...
+  "d0798568-47c3-453e-ae39-242d8a96b1dc": {
+    "collection_id": "Aberrant cGMP Signaling Persists During Recovery in Mice with Oxygen-Induced Pulmonary Hypertension",
+    "community_id": "researchers-collections"
+  },
   "8s45q876k": {
     "collection_id": "Beyond an Apple a Day: Finding Reliable Health Information Online",
     "community_id": "researchers-collections"
   },
+  "5f1b1739-512f-4015-98bc-22f37f42af7b": {
+    "collection_id": "calantone-et-al-2022-commun-biol-source-data",
```
sharpattack, line 88: Double-checking here, because this looks slightly different than the rest. At a glance, all other `collection_id` don't have this format.

Meowcenary: It looks like this entry was just moved up the list from line 422

Meowcenary (approved):
Thanks for adding this @kglibrarian !

dh_collections_prism_collection_community.json
fenekku: I think what @sharpattack means is that for other entries the `collection_id` is not in a slug format but rather a human readable format. So here it would be: 

 ```suggestion
    "collection_id": "Calantone et al 2022 Commun Biol Source Data",
```

Since we are not using collection_ids right now it's not a big deal, but might as well change it.

PR #1125, opened 2022-10-26
[#] Update Memoized Headers

Meowcenary:
Following up from the previous memoized header update. Complex headers and searching vocabularies not typically assigned to a DigitalHub field will be handled in commits separate of this.

---

Update memoized header files to include failed lookups that Gretchen was able to map.


fenekku (approved):
I guess it's fine :wink: 

You know best Eric, if it be valuable now to use a csv or json file to hold these. That way the file can be edited by hand and changes diffed easily.

Meowcenary:
> I guess it's fine 😉
> 
> You know best Eric, if it be valuable now to use a csv or json file to hold these. That way the file can be edited by hand and changes diffed easily.

I should have used json/jsonl from the start, but I think I'm going to leave it as is for the sake of just getting this in to avoid introducing new problems. I can't remember exactly what was going wrong back when I went the csv route, but I do remember having some annoying issues, probably user error in retrospect, with parsing out the entries.

PR #1126, opened 2022-10-27
[#1122] Return Blank Array For Sizes As Last Alternative

### LINKED ISSUE #1122: [Export] Return [] for sizes as last alternative
The export (2022-10-20) exports `"sizes": null` when https://github.com/galterlibrary/digital-repository/blob/master/app/models/invenio_rdm_record_converter.rb#L615 doesn't match any of the conditional branches. InvenioRDM doesn't accept `null`, it only accepts `[]`. Just a matter of returning `[]` in an else branch.

I will make a temporary fix for the export data, but it should be fixed in the export phase for next imports.

Meowcenary:
When the page count for a file is blank return an empty array to be compatible with the Prism importer. closes #1122


Meowcenary:
Very small PR. Hopefully won't raise much discussion.

carsonicator (approved):
Thanks @Meowcenary

PR #1129, opened 2022-11-03
[#1124] Strip Leading and Ending Whitespace From DOI Identifier

### LINKED ISSUE #1124: [Export] strip leading/ending whitespace from DOI identifier
The export generated:

```json
"pids": {
      "doi": {
        "client": "digitalhub",
        "identifier": " 10.1007/s11899-017-0400-3",
        "provider": "datacite"
      }
    },
```

but should generate (notice the removed whitespace in front of the identifier):

```json
"pids": {
      "doi": {
        "client": "digitalhub",
        "identifier": "10.1007/s11899-017-0400-3",
        "provider": "datacite"
      }
    },
```

This only affected 1 record in the 2022-10-20 export though.

Meowcenary:
closes #1124

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -137,13 +137,17 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   end
 
   def invenio_pids(doi)
-    {
-      "doi": {
-        "identifier": doi, # doi is stored in an array
-        "provider": "datacite",
-        "client": "digitalhub"
+    if doi.blank?
+      {}
+    else
+      {
+        "doi": {
+          "identifier": doi.strip,
+          "provider": "datacite",
+          "client": "digitalhub"
+        }
       }
-    }
+    end
   end
 
   def invenio_provenance(proxy_depositor, on_behalf_of)


Meowcenary:
Little baby PR for this small issue 👶 🍼

app/models/invenio_rdm_record_converter.rb
```diff
@@ -139,7 +139,7 @@ def record_for_export
   def invenio_pids(doi)
     {
       "doi": {
-        "identifier": doi, # doi is stored in an array
+        "identifier": doi&.strip,
```
Meowcenary, line 142: Removing the comment since it was from when we first started on this and were not aware of how DigitalHub was storing things.

carsonicator (approved):
Thanks @Meowcenary

fenekku (approved):
See comment but it's not a blocker.

app/models/invenio_rdm_record_converter.rb
Meowcenary: I will make this change quickly. I did not make the change initially thinking that we had intentionally left the logic as is for some reason, but it definitely seemed suspicious.

PR #1130, opened 2022-11-04
[#] Finishing Header Updates

### LINKED ISSUE #1090: Unmapped Subject: Name (LCNAF) terms
I wasn't sure which existing issue to add this to - feel free to move it/delete it as needed.

https://digitalhub.northwestern.edu/files/e342fabd-2faa-42bf-bd39-582fdac49a0d	https://vtfsmghslrepo02.fsm.northwestern.edu/records/9trpe-92w83

have one term in DH that didn't make it to Prism - it's the Subject: Name term "Feinberg School of Medicine. Center for Community Health", which isn't in LCNAF. I understand why this mapping failed, but am unsure what the best solution is. Some thoughts:

1. Can we run a report of all terms in "official" vocabularies in DH that have failed, and change them manually? Some will be typos, but some I think are just entered erroneously and will likely need to be made keywords. This seems to be a better solution in terms of cleaner data, but more time consuming.
2. Is there a way to automatically make a "failed" term a keyword? This seems possibly too complex, but I wasn't sure.

Meowcenary:
* Finished correcting headers where possible
* Convert memoized subject files to JSON to be easier to understand and reason about
* When searching memoized headers all memoized files will be searched. The memoized headers for the searchale vocabulary for a field will be searched first to avoid pids changing for terms that appear in multiple vocabs
* Unmappable complex headers will be exported as keywords rather than matched to a pid from the mesh, lcsh, or lcnaf vocabularies

This will close #1090 , but has some other unrelated work included. The reason it has the name it does is a consequence of how long and meandering this work has been. I can rename the branch if it really bothers anyone.


app/models/header_lookup.rb
```diff
@@ -1,14 +1,14 @@
 class HeaderLookup
-  MEMOIZED_MESH_FILE = "memoized_mesh.txt"
-  MEMOIZED_LCSH_FILE = "memoized_lcsh.txt"
-  MEMOIZED_LCNAF_FILE = "memoized_lcnaf.txt"
+  MEMOIZED_MESH_FILE = "memoized_mesh.json"
```
Meowcenary, line 2: We can probably remove the memoized files that were text formatted as Ruby hashes, but I don't think it makes much of a difference one way or the other.

app/models/header_lookup.rb
```diff
@@ -1,14 +1,14 @@
...
+  MEMOIZED_LCSH_FILE = "memoized_lcsh.json"
+  MEMOIZED_LCNAF_FILE = "memoized_lcnaf.json"
 
   SEARCHABLE_MESH_FILE = "subjects_mesh.jsonl"
   SEARCHABLE_LCSH_FILE = "subjects_lcsh.jsonl"
   # Needs to be in config because file is too large for VCS
   SEARCHABLE_LCNAF_FILE = ENV["SUBJECTS_LCNAF"]
 
-  ABSENT_SUBJECT = :absent
+  ABSENT_SUBJECT = "absent"
```
Meowcenary, line 11: In order for the JSON library to parse correctly we need to use the string `"absent"` and not the symbol `:absent`. It's pretty annoying , but 🤷

app/models/header_lookup.rb
```diff
@@ -17,9 +17,11 @@ def initialize
     @@searchable_lcnaf_file ||= SEARCHABLE_LCNAF_FILE
 
     # these are values that have been previously found from the searchable terms
-    @@memoized_mesh ||= read_memoized_headers(MEMOIZED_MESH_FILE)
-    @@memoized_lcsh ||= read_memoized_headers(MEMOIZED_LCSH_FILE)
-    @@memoized_lcnaf ||= read_memoized_headers(MEMOIZED_LCNAF_FILE)
+    @@all_memoized_terms = {
```
Meowcenary, line 20: This was done so that we could set a search order for the memoized files. I only noticed this because of a spec that failed, but because we're searching **all** memoized headers now there are situations where a term appears in two vocabs E.g mesh and lcsh. This could then lead to a mesh pid being selected instead of an lcsh pid (or vice versa) when we would prefer to have the lcsh pid. There are situations where we will probably have mixed pids, but I think it's best to try and keep things as consistent as we can.

Meowcenary: This is kind of hard to explain so if it's easier to talk about it let me know and we can set up a quick meeting or something.

app/models/header_lookup.rb
```diff
@@ -61,24 +63,31 @@ def pid_lookup_by_field(term="", field="")
...
-  def read_memoized_headers(filepath)
-    begin
-      file = File.read(filepath)
-      # eval will interpret a ruby string passed as an argument
-      # in this case it constructs a hash from the file
-      eval(file)
-    rescue Errno::ENOENT
-      return {}
+  def search_memoized_terms(term, primary_memoized_vocab)
+    secondary_memoized_vocab = @@all_memoized_terms.keys - [primary_memoized_vocab]
```
Meowcenary, line 76: This is so terrible, but it gets the job of ordering the memoized files to search done

app/models/header_lookup.rb
```diff
@@ -61,24 +63,31 @@ def pid_lookup_by_field(term="", field="")
...
+      if term_pid = @@all_memoized_terms[memoized_vocab][term]
+        return term_pid
+      end
     end
+
+    # nothing has been found in ANY memoized file
+    nil
+  end
+
+  def read_memoized_headers(filepath)
```
Meowcenary, line 90: This is not only cleaner, but more secure 🔒 Eval is _very_ dangerous ⚠️

app/models/invenio_rdm_record_converter.rb
```diff
@@ -626,4 +626,16 @@ def sizes
   def user_email(user)
     user.email == "joshelder@northwestern.edu" ? "JoshElder@northwestern.edu" : user.email
   end
+
+  # This is the current list of complex headers that were not mapping correctly. They will be exported as keywords
+  def unmappable_complex_headers
```
Meowcenary, line 631: I dont love hardcoding this, but as I've been saying across this PR - it gets the job done.

Meowcenary:
One of the many issues with the old format for the memoized file was that it was a little difficult to spot duplicates. It gets into semantics, but [here's some discussion on JSON formatting with regards to duplicate keys](https://stackoverflow.com/questions/21832701/does-json-syntax-allow-duplicate-keys-in-an-object).

To be extra sure I just checked the output with `sort <memoized file name > | uniq -d` which did not find any duplicates.

app/models/header_lookup.rb
fenekku: Removing them is good and cleaner. They will still be in version control anyway.

app/models/header_lookup.rb
```diff
@@ -61,24 +63,31 @@ def pid_lookup_by_field(term="", field="")
         end
       end
 
-      memoized_terms[term] = term_pid
-      File.write(memoized_file, memoized_terms)
+      @@all_memoized_terms[memoized_vocab][term] = term_pid
+      File.write(memoized_file, @@all_memoized_terms[memoized_vocab])
```
fenekku, line 67: This should be some form of `JSON.dump`, right? (otherwise we are just dumping the hash again?)

app/models/header_lookup.rb
```diff
@@ -61,24 +63,31 @@ def pid_lookup_by_field(term="", field="")
...
+        return term_pid
+      end
     end
+
+    # nothing has been found in ANY memoized file
+    nil
+  end
+
+  def read_memoized_headers(filepath)
+    file = JSON.parse(File.read(filepath))
```
fenekku, line 91: minor: can probably get rid of `file` especially since it's not a file at this point but the Json hash.

app/models/invenio_rdm_record_converter.rb
fenekku: There is no rule for defining those, eh? It will indeed have to do :shrug:

memoized_person_or_org_data.txt.bak
```diff
@@ -0,0 +1 @@
...personal", "given_name"=>"Jasmine Ashley", "family_name"=>"Lucas"}}, "Laramy, Christine Rose"=>{:person_or_org=>{"type"=>"personal", "given_name"=>"Christine Rose", "family_name"=>"Laramy"}}, "Nandwana, Vikas"=>{:person_or_org=>{"type"=>"personal", "given_name"=>" Vikas", "family_name"=>"Nandwana"}}, "O'Brien, Matthew Nathan"=>{:person_or_org=>{"type"=>"personal", "given_name"=>"Matthew Nathan", "family_name"=>"O'Brien"}}, "Liu, Xiaolong"=>{:person_or_org=>{"type"=>"personal", "given_name"=>"Xiaolong", "family_name"=>"Liu"}}, "Chen, Bor-Rong"=>{:person_or_org=>{"type"=>"personal", "given_name"=>"Bor-Rong", "family_name"=>"Chen"}}, "Crosby, Lawrence Anthony"=>{:person_or_org=>{"type"=>"personal", "given_name"=>"Lawrence Anthony", "family_name"=>"Crosby"}}, "Pastva, Joelen"=>{:person_or_org=>{"type"=>"personal", "given_name"=>" Joelen", "family_name"=>"Pastva", "identifiers"=>[{:scheme=>"orcid", :identifier=>"0000-0001-9772-5144"}]}}, "Miller, Alfred Frederick, Jr."=>{:person_or_org=>{"given_name"=>"Alfred Frederick Jr.", "family_name"=>"Miller", "type"=>"personal"}}, "Mckay, Frederick, S."=>{:person_or_org=>{"given_name"=>"Frederick S.", "family_name"=>"Mckay", "type"=>"personal"}}}
```
fenekku, line 1: Was this supposed to be here?

app/models/header_lookup.rb
Meowcenary: Oh jeez, good catch.

Meowcenary: Thankfully this was as simple as tacking a `.to_json` on

memoized_person_or_org_data.txt.bak
Meowcenary: Nope 😓

carsonicator (approved):
Thanks @Meowcenary

app/models/header_lookup.rb
```diff
@@ -61,24 +63,32 @@ def pid_lookup_by_field(term="", field="")
         end
       end
 
-      memoized_terms[term] = term_pid
-      File.write(memoized_file, memoized_terms)
+      @@all_memoized_terms[memoized_vocab][term] = term_pid
+
+      File.write(memoized_file, @@all_memoized_terms[memoized_vocab].to_json)
```
fenekku, line 68: If we want to edit these files by hand/see diffs better, something like:
https://docs.ruby-lang.org/en/2.7.0/JSON.html#method-i-pretty_generate is probably what we want (otherwise we will get a very long 1 line json object). I leave it as your call.

fenekku (approved):

app/models/header_lookup.rb
Meowcenary: This makes more sense. I actually have never used that method so I'm going to test it a bit to make sure it's not going to do anything unexpected.

Issue #1100, opened 2022-09-21
GV Black item issues 


gneidhardt:
I assume the issues in this document might be found in other GV Black material (especially for one page items). I assume issues come from combining archival items, but this item is titled "...combined" even though it's a 1:1 item match for its DH counterpart; should be part of GV Black papers or subset, but no collection in Prism; DOI is completely; and ARK is present in Prism but not DH:
- https://digitalhub.northwestern.edu/files/7c298e55-bf15-4435-a625-943da8322274
- https://vtfsmghslrepo02.fsm.northwestern.edu/records/wd100-mv020

In another GV Black set of items, the date created is less specific in Prism than in DH; ARK is present in Prism but not DH; no collection in Prism:
- https://digitalhub.northwestern.edu/files/5h73pw137
- https://vtfsmghslrepo02.fsm.northwestern.edu/records/hg1vf-nws30

Another GV Black set of issues - ARK is present in Prism but not DH; no collection in Prism (should be part of GV Black); original identifier present in DH but not Prism:
- https://digitalhub.northwestern.edu/files/c247ds096
- https://vtfsmghslrepo02.fsm.northwestern.edu/records/zja7n-1ch12

And rounding it out with some extra weird GV Black behavior. For this one, I corrected a MeSH term in DH (was a previous version of term that is no longer valid), but I'm unsure if this will be imported next time, or if this one might be weird since it's a GV Black item; LCNAF name in DH didn't transfer to Prism (front or back end); format is wrong (holdover from system-generated data in DH (jpeg listed instead of tiff, which is the actual file format):
- https://digitalhub.northwestern.edu/files/hq37vn595
- https://vtfsmghslrepo02.fsm.northwestern.edu/records/41r2w-ay161

gneidhardt:
Adding @kglibrarian to review above information - it looks like the files that were migrated perhaps weren't the files intended for migration, so I hesitate to edit the "-combined" files listed above in DH if those aren't the ones that are meant for migration.

Meowcenary:
- After the latest import is ran by @fenekku , follow up with @gneidhardt to ensure that none of the records listed in this ticket still have issues
- If there are lingering issues open a new ticket to specify what is going wrong with those particular records

fenekku:
From 2022-10-20 export/import:

| Title | Digitalhub | Prism | Status |
| ----- | ---------- | ----- | ------ |
[...]

PR #1131, opened 2022-11-08
[#1100] GV Black Item Issues

Meowcenary:
Export script will now remove the previous export directory automatically if it exists. This will prevent any situations where files that had previously been exported, but are no longer being exported are still included by accident.

### DIFF
--- a/lib/scripts/repo_export_in_batches.rb
+++ b/lib/scripts/repo_export_in_batches.rb
@@ -1,10 +1,11 @@
-# Before you run this script empty the directory tmp/export/!
-#   rm -rf tmp/export/; mkdir tmp/export/
 # Run this script with the following:
 #   ./bin/rails r lib/scripts/repo_export_in_batches.rb > "$(date +'%Y-%m-%d-%H%M%S')_repo_export_in_batches.log"
-# OR
-#   rm -rf tmp/export/; mkdir tmp/export/; ./bin/rails r lib/scripts/repo_export_in_batches.rb > "$(date +'%Y-%m-%d-%H%M%S')_repo_export_in_batches.log"
-#
+
+puts "---------\nRemoving previous export (if it exists) at tmp/export\n---------"
+export_dir = "tmp/export"
+FileUtils.rm_rf(Dir[export_dir])
+FileUtils.mkdir_p(export_dir)
+
 puts "---------\nBeginning repo export at #{Time.now} #{Time.zone}\n---------"
 
 # set classes to have records exported and classes that will do the actual conversion
@@ -35,8 +36,6 @@ converters.each do |converter|
   conversion_count = 0
 
   converter[:model_class].find_each do |record_for_export|
-    next if record_for_export.is_a?(Page)
-
     converted_record = converter[:converter_class].new(
       record_for_export, collection_store.data, role_store.data
     )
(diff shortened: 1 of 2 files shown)


lib/scripts/repo_export_in_batches.rb
```diff
@@ -4,7 +4,10 @@
 #   ./bin/rails r lib/scripts/repo_export_in_batches.rb > "$(date +'%Y-%m-%d-%H%M%S')_repo_export_in_batches.log"
 # OR
 #   rm -rf tmp/export/; mkdir tmp/export/; ./bin/rails r lib/scripts/repo_export_in_batches.rb > "$(date +'%Y-%m-%d-%H%M%S')_repo_export_in_batches.log"
-#
+
+puts "---------\nRemoving previous export (if it exists) at tmp/export\n---------"
+FileUtils.rm_rf(Dir['tmp/export'])
```
Meowcenary, line 9: I tested this out locally and on staging so I'm confident this will work on production as well.

carsonicator (approved):
Thanks @Meowcenary

lib/scripts/repo_export_in_batches.rb
sharpattack: Does this just delete the contents of `/tmp/export` or will the directory need to be created again? Also, can we update the comments/instructions above?

Meowcenary: Nice catch! For some reason I thought that file writing would just build the path, but I think I'm conflating with something else. Anyways, that's been changed and so have the instructions.

Meowcenary:
After meeting with Karen and Guillaume I've got two things to change:

- The collection "areybook" is going to be added to the unexportable logic
- We are probably going to remove the skipping of Page objects in the top level script, but this needs to be tested first

lib/scripts/repo_export_in_batches.rb
```diff
@@ -1,10 +1,11 @@
...
 # Run this script with the following:
 #   ./bin/rails r lib/scripts/repo_export_in_batches.rb > "$(date +'%Y-%m-%d-%H%M%S')_repo_export_in_batches.log"
-# OR
-#   rm -rf tmp/export/; mkdir tmp/export/; ./bin/rails r lib/scripts/repo_export_in_batches.rb > "$(date +'%Y-%m-%d-%H%M%S')_repo_export_in_batches.log"
-#
+
+puts "---------\nRemoving previous export (if it exists) at tmp/export\n---------"
+export_dir = "tmp/export"
+FileUtils.rm_rf(Dir[export_dir])
+FileUtils.mkdir(export_dir)
```
fenekku, line 7: We are probably safer with `mkdir_p` to create `tmp` if it doesn't exist.

fenekku (approved):

lib/scripts/repo_export_in_batches.rb
Meowcenary: Good point

lib/scripts/repo_export_in_batches.rb
```diff
@@ -35,7 +36,7 @@
   conversion_count = 0
 
   converter[:model_class].find_each do |record_for_export|
-    next if record_for_export.is_a?(Page)
+    # next if record_for_export.is_a?(Page)
```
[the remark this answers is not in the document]
Meowcenary: 🥳

PR #1140, opened 2022-11-17
[#1136] Singularize pages When Just One

### LINKED ISSUE #1136: [Export] Singularize "pages" when just 1
This isn’t exactly a metadata error, but I did notice that one page items still use plural pages (ie “1 pages”). See https://vtfsmghslrepo02.fsm.northwestern.edu/records/hqd4z-5a620. Not sure if this is worth addressing, if it has come up already, or if it’s fine to leave. Just read a little oddly to me, but I can live with it

Meowcenary:
Singularize "pages" to "page" for the metadata field size when there is just one page. closes #1136

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -617,12 +617,15 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   end
 
   def sizes
+    page_count = @generic_file.page_count&.shift
+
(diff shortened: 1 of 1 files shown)


Meowcenary:
I've tested this out in the terminal and it's working as expected. I was going to add some specific specs for this, but mocking out the "sizes" method was proving to be a pain because we assign the file as an instance variable. We _could_ refactor the size method to take page_count as an argument and then testing would be easy, but it's such a small change that it just seems unnecessary given the other tasks we have to get done still.

fenekku (approved):

PR #1148, opened 2022-12-07
[#1123] Titlecase Josh Elder Email Whereever It Is

### LINKED ISSUE #1123: [Export] Titlecase josh's email wherever it is
We've fixed Josh's email in the `owner` field (https://github.com/galterlibrary/digital-repository/pull/1106), but we should have fixed it everywhere instead :facepalm: . The export generates:

e.g., 

```
"extras": {
    "permissions": {
      "edit": {
        "jbe2215": "joshelder@northwestern.edu", 
        "keg827": "karen.gutzman@northwestern.edu"
      }, 
      "owner": {
        "jbe2215": "JoshElder@northwestern.edu"
      }, 
      "read": {
        "public": ""
      }
    }
  }, 
```

We want `"jbe2215": "JoshElder@northwestern.edu"` in the `"edit"` in this case. In general we want that conversion for any permission field.

Meowcenary:
Handle cases where permissions are in role store. closes #1123

### DIFF
--- a/app/models/invenio_rdm_record_converter.rb
+++ b/app/models/invenio_rdm_record_converter.rb
@@ -79,8 +79,8 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
     if !@generic_file.based_near.empty?
       data["presentation_location"] = @generic_file.based_near
     end
-    data["permissions"] = file_permissions
 
+    data["permissions"] = file_permissions
     data
   end
 
@@ -95,15 +95,14 @@ class InvenioRdmRecordConverter < Sufia::Export::Converter
   end
 
   def file_permissions
-    permission_data = Hash.new
-
+    permission_data = {}
     permission_data["owner"] = owner_info(@generic_file.depositor)
 
     @generic_file.permissions.each do |permission|
-      permission_data[permission.access] ||= Hash.new
+      permission_data[permission.access] ||= {}
 
-      if @role_store[permission.agent_name]
-        permission_data[permission.access].merge!(@role_store[permission.agent_name])
+      if role_user_data = @role_store[permission.agent_name]
+        permission_data[permission.access].merge!(role_user_data)
       elsif user = User.find_by(username: permission.agent_name)
         permission_data[permission.access].merge!({user.username => normalize_email(user)})
       else
--- a/app/models/role_store.rb
+++ b/app/models/role_store.rb
@@ -6,8 +6,18 @@ class RoleStore
   end
 
   def build_role_store_data
-    Role.find_each do |r|
-      @data[r.name] = r.users.pluck(:username, :email).to_h
+    josh_elder_username = "jbe2215"
+    josh_elder_email  = "JoshElder@northwestern.edu"
+
+    Role.find_each do |role|
+      role_user_data = r.users.pluck(:username, :email).to_h
+
+      # update Josh's email to titlecase in the role store
+      if role_user_data[josh_elder_username]
+        role_user_data[josh_elder_username] = josh_elder_email
+      end
+
+      @data[role.name] = role_user_data
     end
   end
 end
--- a/memoized_lcnaf.json
+++ b/memoized_lcnaf.json
@@ -140,7 +140,7 @@
   "Black, G. V. (Greene Vardiman), 1836-1915.": "http://id.loc.gov/authorities/names/n80164319",
   "Howe, James Lewis, 1859-": "http://id.loc.gov/authorities/names/no2010017369",
   "Johnson, Charles Nelson, 1860-1938": "http://id.loc.gov/authorities/names/n84805773",
-  "Whitslar, William H., 1862-1930.": "absent",
+  "Whitslar, William H., 1862-1930.": "http://id.loc.gov/authorities/names/n2012189982",
   "Hugenschmidt, Arthur C. (Arthur Christopher), 1862-1929": "http://id.loc.gov/authorities/names/nr2006019532",
   "Newkirk, Garrett, 1847-1921.": "absent",
   "Newkirk, Garrett, 1847-1921": "http://id.loc.gov/authorities/names/n98076155",
--- a/memoized_lcsh.json
+++ b/memoized_lcsh.json
@@ -408,5 +408,7 @@
   "Illinois--Chicago Metropolitan Area": "https://id.loc.gov/authorities/subjects/sh2009008684-781",
   "Latin America": "http://id.loc.gov/authorities/subjects/sh85074879",
   "Viral Proteins": "https://id.loc.gov/authorities/subjects/sh94006884",
-  "Cytology": "https://id.loc.gov/authorities/subjects/sh99005436"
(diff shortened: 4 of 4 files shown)


app/models/invenio_rdm_record_converter.rb
```diff
@@ -79,33 +79,33 @@ def extra_data
...
+      {user.username => user_email(user.email)}
     else
       {"unknown": "unknown"}
     end
   end
 
   def file_permissions
-    permission_data = Hash.new
-
+    permission_data = {}
```
Meowcenary, line 98: I can't remember exactly why I was using `Hash.new` constructor syntax rather than `{}`, but regardless I think this is better.

memoized_lcnaf.json
```diff
@@ -140,7 +140,7 @@
   "Black, G. V. (Greene Vardiman), 1836-1915.": "http://id.loc.gov/authorities/names/n80164319",
   "Howe, James Lewis, 1859-": "http://id.loc.gov/authorities/names/no2010017369",
   "Johnson, Charles Nelson, 1860-1938": "http://id.loc.gov/authorities/names/n84805773",
-  "Whitslar, William H., 1862-1930.": "absent",
+  "Whitslar, William H., 1862-1930.": "http://id.loc.gov/authorities/names/n2012189982",
```
Meowcenary, line 143: I'm not sure why this wasn't committed previously. Throwing it in here.

memoized_lcsh.json
```diff
@@ -408,5 +408,7 @@
   "Illinois--Chicago Metropolitan Area": "https://id.loc.gov/authorities/subjects/sh2009008684-781",
   "Latin America": "http://id.loc.gov/authorities/subjects/sh85074879",
   "Viral Proteins": "https://id.loc.gov/authorities/subjects/sh94006884",
-  "Cytology": "https://id.loc.gov/authorities/subjects/sh99005436"
+  "Cytology": "https://id.loc.gov/authorities/subjects/sh99005436",
```
Meowcenary, line 411: Same as `memoized_lcnaf`. Not sure why this wasn't committed, but it seems fine to just throw it in here.

app/models/invenio_rdm_record_converter.rb
```diff
@@ -79,33 +79,33 @@ def extra_data
...
     permission_data["owner"] = owner_info(@generic_file.depositor)
 
     @generic_file.permissions.each do |permission|
-      permission_data[permission.access] ||= Hash.new
+      permission_data[permission.access] ||= {}
 
-      if @role_store[permission.agent_name]
-        permission_data[permission.access].merge!(@role_store[permission.agent_name])
+      if role_user_data = @role_store[permission.agent_name]
+        role_user_data.update(role_user_data) { |user_id, email| user_email(email) }
```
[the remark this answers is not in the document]
fenekku: So I can follow: what is the shape of `@role_store`? And would it be simpler at this point to modify `https://github.com/galterlibrary/digital-repository/blob/57c85146b6853f71f39490363f6974e5ca77fc31/app/models/role_store.rb#L8` to have the joshelder change in there directly at this point?

app/models/invenio_rdm_record_converter.rb
```diff
@@ -631,8 +631,8 @@ def sizes
     end
   end
 
-  def user_email(user)
-    user.email == "joshelder@northwestern.edu" ? "JoshElder@northwestern.edu" : user.email
+  def user_email(email)
```
fenekku, line 634: might as well rename `user_email` to `normalize_email` or some such since `user` isn't related anymore.

Meowcenary: Good point

mattbaumann1: agreed

mattbaumann1 (approved):

app/models/invenio_rdm_record_converter.rb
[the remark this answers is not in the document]
Meowcenary: > what is the shape of `@role_store`?
Not sure exactly what you mean, but it's a hash with a role's name as the key and a value that is itself a hash with a user's NWU NetId as the key and a user's email as the value.

In code:
```
{
  role_name: {
    net_id: "user_email",
    another_net_id: "another_user_email"
  },
  another_role_name {
    yet_net_id:  "yet_another_user_email"
  }
}
```

I thought about making the change at the line you're pointing to, but I figured since we already have the function in the  converter it would be easiest to just reuse it there. I don't have strong opinions so if you would rather I move it to the `role_store` model that works for me.

carsonicator (approved):
Thanks @Meowcenary

app/models/invenio_rdm_record_converter.rb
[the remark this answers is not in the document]
fenekku: Thanks. It's not a big inclination, but I think having `normalize_email` in some `utils.rb` and reusing it in this file and in https://github.com/galterlibrary/digital-repository/blob/57c85146b6853f71f39490363f6974e5ca77fc31/app/models/role_store.rb#L8 file would be good. 

The current update pattern is strange to me: why would we update in-place `@role_store` if we always update it in-place? (and we could do that just once). With the troubles we've been having with this, I would rather we sidestep quirks if we can. I'll approve in advance though.

fenekku (approved):
See comment, good to merge after addressing it.

app/models/invenio_rdm_record_converter.rb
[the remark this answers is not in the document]
sharpattack: Not to stir the pot, but after role_store creation, I would have updated josh elder's email before the export.
https://github.com/galterlibrary/digital-repository/blob/master/lib/scripts/repo_export_in_batches.rb#L30

app/models/role_store.rb
```diff
@@ -6,8 +6,18 @@ def initialize
...
+
+    Role.find_each do |role|
+      role_user_data = r.users.pluck(:username, :email).to_h
+
+      # update Josh's email to titlecase in the role store
+      if role_user_data[josh_elder_username]
+        role_user_data[josh_elder_username] = josh_elder_email
+      end
+
+      @data[role.name] = role_user_data
```
Meowcenary, line 20: Move the value of `r.users.pluck...` into a variable `role_user_data`, check if Josh's netid (username) is in the `role_user_data` and if it is update it to titlecase, store the value of `role_user_data`

app/models/invenio_rdm_record_converter.rb
```diff
@@ -79,33 +79,32 @@ def extra_data
...
+    permission_data = {}
     permission_data["owner"] = owner_info(@generic_file.depositor)
 
     @generic_file.permissions.each do |permission|
-      permission_data[permission.access] ||= Hash.new
+      permission_data[permission.access] ||= {}
 
-      if @role_store[permission.agent_name]
-        permission_data[permission.access].merge!(@role_store[permission.agent_name])
+      if role_user_data = @role_store[permission.agent_name]
```
Meowcenary, line 104: I kind of like having this in a variable so I'm going to keep that change.

PR #1154, opened 2022-12-15
1.30.6 Hotfix For December 2022 Export Updates

Meowcenary:
There was a variable that was misnamed within RoleStore#build_store_data that has been corrected and confirmed working with a small test.


mattbaumann1 (approved):
simple and to the point.

fenekku (approved):

PR #1165, opened 2023-03-16
1.30.7 Update Memoized Vocabulary Terms

Meowcenary:
Add latest memoized terms and update existing memoized terms to map to the correct values where necessary.


carsonicator (approved):
Thanks @Meowcenary

PR #1174, opened 2023-04-03
1117 disable file editing for all except for specific users

Meowcenary:
Disable file editing for anyone that does not have the "sunset-editor" role. Add script to add "sunset-editor" role to specified users. Disable uploading files for everyone.


app/views/roles/edit.html.erb
```diff
@@ -1,26 +1,8 @@
...
-    </li>
+    <li id=<%= user.user_key %>><%= user.user_key %></li>
   <% end %>
 </ul>
-<h3>Add New Account</h3>
-<%= bootstrap_form_tag :url=> role_management.role_users_path(@role) do |f| %>
-  <%= f.text_field 'user_key', :label=>"User" %>
-  <%= f.submit "Add" %>
-  <%= link_to "Cancel", role_management.roles_path, :class => 'btn btn-default' %>
-<% end %>
```
sharpattack, line 26: Why changes to this view?

sharpattack (commented):
I like the new `sunset-editor` role. Can we also add it to `ability.rb`? This should prevent users from trying to visit `/edit` for their files and collections.

mattbaumann1 (approved):

app/views/_user_util_links.html.erb
```diff
@@ -9,14 +9,13 @@
...
       <ul class="dropdown-menu dropdown-menu-right" role="menu">
-        <% if can? :create, GenericFile %>
-          <li><%= link_to raw('<span class="glyphicon glyphicon-upload"></span> upload'), sufia.new_generic_file_path, id: 'upload_nav_link' %></li>
-        <% end %>
         <li><%= link_to raw('<span class="glyphicon glyphicon-dashboard"></span> my dashboard'), sufia.dashboard_index_path, id: 'dashboard_nav_link' %></li>
         <li><%= link_to raw('<span class="glyphicon glyphicon-user"></span> my profile'), sufia.profile_path(current_user), id: 'profile_nav_link' %></li>
-        <%= render partial: 'users/user_util_links_extra' %>
-        <% if can? :manage, 'anything' %>
-          <li><%= link_to raw('<span class="glyphicon glyphicon-king"></span> administration'), role_management.roles_path, id: 'logout_nav_link' %></li>
+        <% if current_user.sunset_editor? %>
```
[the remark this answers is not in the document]
Meowcenary: I don't like that this is what we have to do. It's pretty clearly a hacky solution, but this is a remarkably hacky app. Ideally this would be something handled on a controller, but it's not realistic for the timeline we have and what has to be done.

app/views/roles/edit.html.erb
Meowcenary: It's been so long since I submitted the old PR, but if I remember right it was to stop people from adding or updating roles for themselves. Once this is in we should only be editing a few files. Might be overkill, but it doesn't hurt anything either.

PR #1177, opened 2023-04-12
Hotfix - No Method Error When Current User Nil

Meowcenary:
Calling User#sunset_editor? on nil users raises errors. Add safe navigation operator to calls


fenekku (approved):
Works for me.

